PR #30, opened 2023-02-23
Update to current version on-premise

nomssi:
refactoring of the trans_ref_tab_2_tab( ) method signature, so I am submitting all changes for review


larshp:
CHAR01 is not released in steampunk, http://abapedia.org/steampunk-2302-api/

can `TYPE c LENGTH 1` be used instead?

oblomov-dev:
yes just normal string seems also to be no problem, i merged it and fixed it, it should work now. thanks!

PR #50, opened 2023-02-28
use builtin i instead of ddic int4

larshp:


oblomov-dev:
thanks!

oblomov-dev:
image
merged everything together and in addition also tried to cleanup all dependencies today. now abap2ui5 only uses the above artifacts. that looks v702 ready for me. (ignoring the demo apps) abap2ui5 only consists of the handler class and the draft table. so i now only have to to fix the around 150 syntax issues and then it could work! 😅
@larshp if you know any further problems let me know!

PR #70, opened 2023-03-04
adding logic for typekind_struct1

sun-tassein:
When using a flat structure for the screen for example:
image
no binding is happening for any of the fields.

I tracked the issue to the method get_t_attri_by_ref in local class z2ui5_lcl_utility of class Z2UI5_CL_HTTP_HANDLER change in the PR, the dynamic parser was only working for cl_abap_classdescr=>typekind_struct2 which is deep structure, but not for cl_abap_classdescr=>typekind_struct1 which is flat structure.


oblomov-dev:
thanks a lot for debugging and finding the problem! somehow i missed that.

PR #91, opened 2023-03-14
Downport of feature request: Text area - control cursor position

nomssi:
Work was needed specially on the local classes of Z2UI5_CL_HTTP_HANDLER

The code as is would not compile and a I added local interface Z2UI5_LIF_SYSTEM_RUNTIME to avoid a cyclic dependency on 7.5x.

Some helper methods were added.
The logic of trans_ref_tab_2_tab( ) was slightly changed.
_get_t_attri( ) now recursively calls itself for substructures.
I decided to rename the attribute KIND of structure TY_ATTRI to TYPE_KIND, to enable usage of the CORRRESPONDING operator
An alias was introduced to avoid using the long method form z2ui5_if_view~_generic( ), instead we can write _generic( ).
get_prev_when_no_handler( ) had the same logic twice.

While deploying on the BTP public cloud, I have to update older code version generated by other users. 
Maybe the installation guide could provide a roadmap about how to copy the classes to have different version of the code running in parallel. But it might also be a good thing that only one valid version exist, so it will probably be kept up-to-date? but then abapGit can't be used.

I am still not sure how to select the editor/text area when I want to trigger a scroll down.

OK. That is it for now..


oblomov-dev:
Thanks a lot!

Yes the BTP ABAP Public Trial is difficult to handle, when different users want to use different versions, i think there is no elegant way how wo handle this, but we will see.

I check the changes now and keep a look to the focus again.

PR #96, opened 2023-03-17
`io_app` used dynamically

larshp:
set needed, https://rules.abaplint.org/unused_variables/


oblomov-dev:
i set the severity of abaplint checks only down to warning all the time, but forgot that pragmas are also helpful. thanks for the hint!

larshp:
you want to keep main completely clean, or else which rules should be respected?

if I want to contribute something, first thing is that I have to clean up the code, it should not be that way

larshp:
exactly

on warnings: the severity is mainly a feature because it is part of code inspector

but, if there is a warning, the developer will not know if it is something that is to be enforced or not. Ie. better to have it required, or disable the rule.

oblomov-dev:
ah ok, good to know. so warning is more something from the SAP and atc check world but in a serious ci/cd process it is not really useful. ok i am learning and will keep this in mind :)

PR #106, opened 2023-03-19
fix double space findings

larshp:
and format json file


oblomov-dev:
thanks, i also deactivated double space finding now.

PR #111, opened 2023-03-20
add script setup for running downport logic

larshp:
prerequsites: Node 16 & Bash installed locally

run with `npm install && npm run downport`


oblomov-dev:
thanks a lot! will try this out!

larshp:
still a few bugs, but getting there

PR #135, opened 2023-03-31
Avoid diff for package

### LINKED ISSUE #133: Diffs for on-prem
There are minor diffs when pulling (main or _dev) into a on-prem system (7.54 ABAP Dev Edition):

image

The ABAP language version can't be avoided until we solve https://github.com/abapGit/abapGit/issues/6154.

The package xml is missing the BOM mark. I can open a PR with the correct file if you like.

mbtools:
Adds BOM to `package.devc.xml` to avoid diff for on-prem

Closes #133

PS: If you need to create an XML for another object, it's best to start with a copy of an existing XML-file and upload it to GH. When you edit it afterwards, the format (BOM) will be retained.

### DIFF
--- a/src/00/package.devc.xml
+++ b/src/00/package.devc.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <abapGit version="v1.0.0" serializer="LCL_OBJECT_DEVC" serializer_version="v1.0.0">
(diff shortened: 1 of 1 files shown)


oblomov-dev:
Thank you a lot!

`PS: If you need to create an XML for another object, it's best to start with a copy of an existing XML-file and upload it to GH. When you edit it afterwards, the format (BOM) will be retained.`
Ah this is a very good idea, i will keep this in mind!

PR #161, opened 2023-04-11
fix-syntax-issue-in-older-releases

Trist69:
```like ref to value``` which is ref to data is not a valid syntax in older releases ( <= 7.40 )


oblomov-dev:
thank you!

PR #300, opened 2023-06-22
New TreeTable control

axelmohnen:
New TreeTable control
API Reference: sap.ui.table.TreeTable


oblomov-dev:
Thank you a lot for the contribution! Fell free to share your controls when you extend the view class in the future. Great support!

PR #307, opened 2023-06-23
use ui5 resources with active cachebuster

abs-lme:
Hi, I modified the UI5 CDN URL so it uses the CDN's cache buster.
This makes sure that the UI5 libraries are properly cached and their versions always match (so the used version of `sap.m` is always the same as the version of `sap.ui.core`, which might not be the case if the libraries are loaded without cachebuster).

See UI5 documentation: [Cache Buster for SAPUI5](https://ui5.sap.com/#/topic/91f080966f4d1014b6dd926db0e91070)


oblomov-dev:
Nice, did not know this, but sounds very good! I merge it. Thank you!

PR #313, opened 2023-06-27
TreeTable: new parameters

axelmohnen:
API Reference:
sap.ui.table.TreeTable

New parameters for TreeTable and TreeColumn


oblomov-dev:
thank you

PR #342, opened 2023-07-05
add maxSuggestionWidth to input method

WegnerDan:
Property maxSuggestionWidth of class sap.m.Input added to abap method z2ui5_cl_xml_view->input


oblomov-dev:
great, thank you!

PR #378, opened 2023-07-13
add sap.m.Tree & sap.m.StandardTreeItem

WegnerDan:
add new controls and a way to work with deep tables


oblomov-dev:
Thank you a lot!

Are you on twitter? maybe i will find time in the next days to make a tweet about it. let me know that i can refer to you (just if you want).

WegnerDan:
> Are you on twitter?

I am but I'm not really using it, so no need to refer to me. Thanks!

PR #380, opened 2023-07-13
UI Element Checkbox - Event Select

Th0masMa:


oblomov-dev:
Hi Thomas,

thanks again for the contribution!

Somehow it seems that the whole definition part of the class is overwritten when you send a PR, instead of only add your new lines:
image

Maybe you can check if you use the source code based editor in se80 (or eclipse) when you prepare a PR. I think then only the changed lines will be updated in the PR and i can easily check and merge the changes.

Best regards!

Th0masMa:
Hi oblomov,

I will use the code based editor in the future.

Best Regards
Thomas


Am Do., 13. Juli 2023 um 21:51 Uhr schrieb oblomov ***@***.***

PR #413, opened 2023-07-24
view update - export to xlsx - first gantt compone

choper725:
added support sap.m.table export to xlsx

new gantt chart components added (work in progress for the demo)


oblomov-dev:
Thank you!

PR #427, opened 2023-07-27
New UI Element RatingIndicator

Th0masMa:


oblomov-dev:
nice thank you!

oblomov-dev:
@Th0masMa fyi:
changed the name to range_indicator to make all method names defined in snake_case.
https://github.com/abap2UI5/abap2UI5/blob/ef93a0be94c27c2008e57a6645f9a326a0713e66/src/z2ui5_cl_xml_view.clas.abap#L1361

And changed the input parameters to always clike. When you use boolean it is working for direct values, but you can not use it with a data binding.

PR #433, opened 2023-07-28
added generic way creating date object in XML view

choper725:
added generic way creating date object in XML view

1. added onInit in controller
2. added Date.createObject global symbol


oblomov-dev:
Great, thank you!

PR #464, opened 2023-08-06
Update README.md

oblomov-dev:


larshp:
update https://github.com/abap2UI5/abap2UI5/blob/main/package.json#L22 to latest

oblomov-dev:
Nice thank you!

But is this caused by the update then? Basic rtti_classdescr - seems there is something not correct
image

oblomov-dev:
great! thank you!

PR #468, opened 2023-08-07
added register mesagemanager

choper725:
added register mesagemanager for data validation via binding,

check demo 67, and insert charcters in the Input fields


oblomov-dev:
Nice:
image

Did not know this, thank you.

choper725:
this works on path binding with type Integer and Currency out of the box, which is nice..

PR #509, opened 2023-08-25
add 11 controls to xml view 

choper725:
add 11 controls to xml view


choper725:
hi @oblomov-dev 

what is the error in this PR ? i cant seem to find what it is failing on test

oblomov-dev:
I am not completely sure, but i think the unit test runner on node.js gets confused because there is an abap variable called "for" which is a reserved word in javascript:
image
https://github.com/open-abap/open-abap-core/issues/764

Maybe it will be fixed, but also open-abap is open source so you don't know if this has priority. So as a work around:
You can rename the variable "for" exceptionally maybe to "for_" or "for2". It is not nice but i think a good compromise.

oblomov-dev:
Thank you for the PR @choper725 !

Changed the variable back to "for" sometimes the solution is easier than expected, thank you @larshp for the quick response: https://github.com/open-abap/open-abap-core/issues/764#issuecomment-1693069222

choper725:
no problem @oblomov-dev 
ill PR demos for all of the new control soon

PR #522, opened 2023-09-01
fix zcc_generate_ddic_shlp

choper725:
added ircontroller parameter to method in order to get the current app object


oblomov-dev:
Overlooked this, thank you, added pragma ##NEEDED in the method definition that the code check abaplint knows that the variable "ircontroller" is used inside the method.

choper725:
great!
i updated the demo also with the missing variables

PR #523, opened 2023-09-02
fix generate shlp + messagepopover

choper725:


oblomov-dev:
Thanks! Made the call of the function module dynamic that it can be installed also on abap_cloud.
image

The follwowing type is not always on every abap system:
shlp_descr

It is not found in a cloud system, so it will create a syntax error:
image

Can you copy the type of shlp_descr to a local type and define the variable ls_shlp with this local type instead?

choper725:
ill try,
maybe ill use your help making it to work properly

choper725:
[shlp_descr.txt](https://github.com/abap2UI5/abap2UI5/files/12504439/shlp_descr.txt)

here is my se38 test case for this...

cant seem to get it to work properly.. not sure even why..

choper725:
never mind, got it to work.. ill update this pull request

oblomov-dev:
thanks for the adjustments!

i added some fixes and there is now just one problem left:
image

Do you now why there are type problems here and can you fix this?
Seems that a structure is written into a table.

choper725:
ill check it, maybe i missed something adjusting the new types..

choper725:
image

the FS - <fs_isshlp_fielddescr>
is assigned to the table FIELDDESCR from the structure LS_SHLP,
its actually a table that is getting assigned then...

oblomov-dev:
I separated the definition of the field-symbol now and defined it as a standard table. Did not check the functionality, but hope it still works.

PR #524, opened 2023-09-03
fix zfc_ddic_search_help TT declerations

choper725:
not working with EMPTY KEY decleration in TT of method zfc_ddic_search_help

changed to DEFAULT KEY


oblomov-dev:
thank you!

PR #545, opened 2023-09-11
added doctype tag to http handler + rich text edit

choper725:
added <!DOCTYPE html> to http handler class
in order to support rich text editor TinyMCE.

add rich text editor + demo 106


oblomov-dev:
thank you!

PR #560, opened 2023-09-22
MultiComboBox fix

noso2k1:
Method multi_combobox was using the generic "ComboBox" instead of "MultiComboBox".


oblomov-dev:
Thank you!

PR #605, opened 2023-10-28
font awsome free icons support

choper725:
will make a guide how the resources should be handled..


oblomov-dev:
cool 😎 thanks!

choper725:
it got also the js animation option and full support for pro version.

needs to tweak the metadata file in order for it to work but for the free version now its fully compatible, with the jadeliver resources i created from github

oblomov-dev:
yes good work and great idea 👍

image

oblomov-dev:
@choper725  fyi reassigned the font class to package 01

PR #607, opened 2023-10-30
rename font awesome + implement driver js

choper725:
hi @oblomov-dev

1. renamed font awsome to font awesome
2. implemented driver js library (awsome feature! - check demo 119 - press TOUR )
3. added NEST_NAV_CONTAINER_TO & NEST2_NAV_CONTAINER_TO & DRIVE to event_frontend to the CASE in http_handler class.

implementation of driverjs is still not fully finished, will finish it in the upcoming days


oblomov-dev:
Nice thank you! Adding this library is a great idea. I will check the demo. 

FYI Needed to do some small abaplint fixes.

larshp:
note the change https://github.com/abap2UI5/abap2UI5/blob/main/.abapgit.xml#L17-L18, it should not be nessesary

oblomov-dev:
@larshp thanks for the hint, did not see this, deleted the classes manually. will also delete these lines in the abapgit.xml

oblomov-dev:
@choper725 reassigned the driver class and did some renaming.

PR #612, opened 2023-11-01
update driverjs

choper725:


oblomov-dev:
@choper725 added abaplint fixes
image

PR #614, opened 2023-11-01
finish driverjs implementation

choper725:


oblomov-dev:
thank you! added some fixes that all tests are passing.

PR #622, opened 2023-11-03
driverjs update

choper725:
now support js events as parameters

added highlight function.

performence & logic fixes


oblomov-dev:
thanks!

PR #636, opened 2023-11-08
Cross Application Navigation to external window

axelmohnen:
If the optional parameter ARGS[3] is equal "EXT" than the new app will be called in new window (browser tab)


oblomov-dev:
thanks!

Issue #496, opened 2023-08-16
Modify JSONModel (best practice?)


WegnerDan:
I have some suggestionitems that are a couple more than 100 items.

I've googled arount a bit and found out there is a default limit of 100 items for suggestionItems in the JSONModel.

I tested increasing the limit and it worked, but I had to modify a class of abap2ui5:
https://github.com/abap2UI5/abap2UI5/commit/5878de9dd6bf843889f0e5de8c2190760b15d213

So my question is, what would be best practice here? Increasing the limit for all abap2ui5 apps doesn't seem right to me.

Is there a way to modify the model without modifying ABAP class z2ui5_cl_http_handler? 
I don't know enough javascript, but maybe the model can be modified by adding some more js after the standard js? Maybe a method to include custom js would help, but I'm not sure.

oblomov-dev:
Hi @WegnerDan,

yes extending the Model to allow more entries is one option, but I am not sure if this is generally the best way.

It is no problem to add custom JS on app level, you can check the examples for this, but i will try to keep the initial index.html as small as possible. But yes maybe in the future there should be the option to extend also there custom JS. I will check this and i can maybe add an importing parameter to the initial get request where JS can be added.

Concerning your problem:
I added a few weeks ago the option to update model data only. Maybe when you have a suggestion table with a lot of entries it is the best way to make filtering on the server side and update the suggestion table depending on the input. Added the following example:
z2ui5_cl_app_demo_60
<img width="600" alt="image" src="https://github.com/abap2UI5/abap2UI5/assets/102328295/38919a24-23a5-4dd1-be67-48921a4c08a1">
There the suggestion table can have a lot of entries but only the possible entries are send to the frontend dynamically.

Does this help you or what are the specific controls you use? Feel free to share you code snippet.

Best regards.

oblomov-dev:
change the model entry limit is now importing parameter of the http_get method. i close this issue, just open a new one if you have further problems.
Check https://github.com/abap2UI5/abap2UI5/pull/681

WegnerDan:
Thanks!

PR #681, opened 2023-11-26
json model size limit parameter to http handler

choper725:
added setSizeLimit to json model as a parameter of HTTP_GET method.. default to 100 (as SAPUI5 Model default).

for now, used in oView, oViewNest, oViewNest2

need to add oViewPopup & oViewPopover (&fragment)

as continue to #496


oblomov-dev:
@choper725 
two changes:
* added the size limit for popup and popover.
* changed the value to a global variable, which makes the maintenance of the index.html easier (have to copy&paste the JS from VS Code to abap class from time to time)
https://github.com/abap2UI5/abap2UI5/blob/77201b6aa249f28d4fd24fbabd6b17605d6051e2/src/z2ui5_cl_fw_http_handler.clas.abap#L583

PR #694, opened 2023-11-30
bugfix stringtable

oblomov-dev:


ssocratous71:
Excellent!!! yes it works fine now!!! You have faster and better support than SAP itself :)!!!

PR #700, opened 2023-12-01
xml view add/update controls

choper725:


oblomov-dev:
nice, as always thank you!

PR #749, opened 2023-12-29
debugging tools fixed LTR of dialog

choper725:


oblomov-dev:
Thank you!

Made some small adjustments.

(1)
To avoid future incompatibilities the use of sap.ui.getCore is not recommended, therefore i changed it to this:
[              `           $( oFragmentController.oDialog .getDomRef() ).css("direction","LTR");` && |\n|  &&](https://github.com/abap2UI5/abap2UI5/blob/ce7a95728df8e20e1a2d3e2e32923137a44e58c1/src/01/03/z2ui5_cl_fw_cc_debugging_tools.clas.abap#L196)

(2) The use of jquery is also not recommended anymore, is there a chance to replace $?
https://blogs.sap.com/2023/11/06/get-rid-of-jquery-in-ui5/

(3) Changed the shortcut to ctrl+f12, good hint 👍

(4) I can not send the XML of the popop as it is to the frontend, otherwise i will get an error when starting from a launchpad with:
https://github.com/abap2UI5/abap2UI5-launchpad_on_premise
Therefore i changed it to this:
https://github.com/abap2UI5/abap2UI5/blob/ce7a95728df8e20e1a2d3e2e32923137a44e58c1/src/01/03/z2ui5_cl_fw_cc_debugging_tools.clas.abap#L169
And yes it is not nice, just quick and dirty solution, will try to find a better way.

Best regards.

PR #751, opened 2023-12-30
debugging tools - LTR

choper725:
1. added predefined css class in http handler.
2. removed jQuery from debugging tools, and made use of predefined css


oblomov-dev:
thank you!

PR #754, opened 2024-01-01
update cjs

choper725:


choper725:
i need to implement a better code for this, for now this is only for checking and see what abap2ui5 is capable in doing with this..
as this is a pure js -json complex library..
might need to write some CC in order to handle the script loading, as it overlapps eachother and creates issues..

oblomov-dev:
thank you!

yes its complex, I also just try around a lot till i get a final and "good enough" solution.

the boolean handling between ABAP and JS will never be perfect but /UI2/cl_json does a good job, so maybe working with a custom control and defining boolean attributes helps.

Let me know if you run into problems.

choper725:
there are chartjs properties that can recive values like:
true, false, ‘+1’ , 3, array of string

in all colora of the rainbow

oblomov-dev:
ah ok, yes is not easy to handle. /ui2/json in general can do it depending on your typing and data structure but sometimes you get not the exact result you need, spend also a lot of time with these problems.

choper725:
well compremises will be made in the conversion of it,
plan b will be ALOT of work building a json string from a abap table in pure js, 
any way a custom control for loading js libraries need to be done in order to avoid overhead in the framework regarding loading multiple js file async, it seems not to behave like a plain html script loading

PR #778, opened 2024-01-12
update chartjs CC

choper725:


choper725:
hi @oblomov-dev 

thx for the tip, made it to work,
but now i get a new abap error updating the model,

image


i updated demo 148 with this error

oblomov-dev:
Thank you I will check this.

choper725:
> Thank you I will check this.

found a bug in the new model update fix you made, (you said expect bugs :) ),
before a model update the binded model and the object (chartjs object model) is the same

after a model update it seems that not all values are getting copied to the model
image

at first before model update everything is fine..

still trying to understand what causes this..

oblomov-dev:
to the first problem, i can see the error but don't know what the initial problem is. I am wondering about the following:
plugins is type string_table and it is empty with the first request:
image

but at the forntend, values are filled:
image

How are you doing this? ;)

i think the problem is then just a follow up error because, the binding method wants to write a table with multiple colums to a string table this creates the error here:
image

i iwll check the second error after fixing the first one, maybe they correspond to each other.

choper725:
haha yes you are right, the plugins is an object insert via the CC render based on the js libraries you want to load in the load_js method.
but i think im having problems (the second problem), even when there are no plugins choosen.

ill create a demo for it without plugins.
the pictures from abap debugger are from that demo, ill just PR it.

oblomov-dev:
(1)
ok then i think it is not the problem with the binding. just try to delete the attributes again before the backend request, otherwise you will map a table with multiple rows into a string_table (plugins) and this can not work.

(2)
yes a demo just with the specific issue would be nice.

choper725:
updated demo 148 with the problem,
the abap debugger pictures from above are relevant

oblomov-dev:
lets continue here https://github.com/abap2UI5/abap2UI5/issues/782

PR #779, opened 2024-01-13
Fix trimming of horizontal tabs

christianguenter2:
Before:
grafik
grafik

After:
grafik


oblomov-dev:
nice, already saw this bug, but too lazy for searching the error, so thank you for fixing :)

PR #796, opened 2024-01-19
update

choper725:


oblomov-dev (commented):
Is ty_chart needed?

choper725:
this type is used when you build a chart thats why the error not used in the cc class

src/02/z2ui5_cl_cc_chartjs.clas.abap
```diff
@@ -492,15 +502,15 @@ CLASS z2ui5_cl_cc_chartjs DEFINITION
     TYPES:
       BEGIN OF ty_chart,
```
larshp, line 503:
```suggestion
      BEGIN OF ty_chart ##NEEDED,
```

oblomov-dev:
@choper725 next time you can fix the unused issue with the ##NEEDED pragma:
image

Thank you as always for your changes, just fyi, i get an ID error in demo 148:
image

src/02/z2ui5_cl_cc_chartjs.clas.abap
oblomov-dev: thanks!

PR #802, opened 2024-01-21
Popups   draft handling   utility

oblomov-dev:


oblomov-dev:
mhh 🤔
image

oblomov-dev:
checked it again and deleted class z2ui5_cl_popup_to_select, because this was the only main change i did, but downport sill runs into problems, mhh no idea why...

PR #803, opened 2024-01-21
add annotations to chartjs

choper725:


choper725:
also changed xfeld field to abap_bool, and modified to demo with the new compress syntax

oblomov-dev:
nice thank you!

PR #825, opened 2024-01-26
improve POPUP_TO_SELECT

christianguenter2:
- add parameter for title
- use medium DDIC description if available

before
grafik

after
grafik

### DIFF
--- a/src/02/03/z2ui5_cl_popup_to_select.clas.abap
+++ b/src/02/03/z2ui5_cl_popup_to_select.clas.abap
@@ -9,6 +9,7 @@ CLASS z2ui5_cl_popup_to_select DEFINITION
     CLASS-METHODS factory
       IMPORTING
         i_tab           TYPE STANDARD TABLE
+        i_title         TYPE clike OPTIONAL
       RETURNING
         VALUE(r_result) TYPE REF TO z2ui5_cl_popup_to_select.
 
@@ -31,6 +32,7 @@ CLASS z2ui5_cl_popup_to_select DEFINITION
     DATA check_initialized TYPE abap_bool.
     DATA check_table_line TYPE abap_bool.
     DATA client TYPE REF TO z2ui5_if_client.
+    DATA title TYPE string.
     METHODS on_event.
     METHODS display.
     METHODS set_output_table.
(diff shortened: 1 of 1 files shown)


christianguenter2:
hm, I'm not sure what to do about the failing tests.

src/02/03/z2ui5_cl_popup_to_select.clas.abap
```diff
@@ -81,7 +85,12 @@ CLASS z2ui5_cl_popup_to_select IMPLEMENTATION.
 
     DATA(columns) = tab->columns( ).
     LOOP AT lt_comp INTO ls_comp.
-      columns->column( width = '8rem' )->header( ns = `` )->text( text = ls_comp-name ).
+      DATA(text) = COND #(
+                     LET scrtext_m = CAST cl_abap_elemdescr( ls_comp-type )->get_ddic_field( )-scrtext_m IN
```
larshp, line 89: as far as I remember, the get_ddic_field() does not work on steampunk/abap cloud programming model

oblomov-dev: yes, maybe you can try to read it in the steampuncification way:
https://github.com/heliconialabs/steampunkification

Or just use the shortcut:
Make dynamic method calls out of it, then the check is also correct.

I know it is annoying, but i think its better to keep the project runnable on both language versions.

And the improvement is a nice idea, thank you for this!

larshp: note: the method exists, it just shortdumps if I remember correctly

so maybe need to do https://github.com/heliconialabs/steampunkification?tab=readme-ov-file#find-abap-language-version-for-running-program before calling it

christianguenter2: Thanks for the hints. I'll try to come up with steampunk compatible solution. 
Of course it's important that this great project runs on all ABAP versions.

christianguenter2: Crafted a solution which runs (hopefully) on all ABAP versions.

src/01/01/z2ui5_cl_util_func.clas.abap
```diff
@@ -1680,4 +1694,89 @@ CLASS Z2UI5_CL_UTIL_FUNC IMPLEMENTATION.
     RAISE EXCEPTION TYPE z2ui5_cx_util_error EXPORTING val = v.
 
   ENDMETHOD.
+
+
+  METHOD rtti_get_data_element_texts.
```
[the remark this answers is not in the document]
oblomov-dev: 😍

oblomov-dev:
thank you, very nice feature!

PR #826, opened 2024-01-26
text element on new popup classes + range tokens

choper725:
+ xml additions


choper725:
so long to cross check…

oblomov-dev:
The Cross Check is not important for merging, you have to update the branch, I made updates in between

choper725:
> The Cross Check is not important for merging, you have to update the branch, I made updates in between

still takes hours.. is it a problem ?

oblomov-dev:
No it's no problem, but did you see my comment concerning the text elements you created for the range values? I am not sure if it's a good idea merging this. I think selects will not work when you change the value of LE, GE etc

choper725:
> No it's no problem, but did you see my comment concerning the text elements you created for the range values? I am not sure if it's a good idea merging this. I think selects will not work when you change the value of LE, GE etc

the value is not text-element only the names.. will this cause a problem ??

oblomov-dev:
yes this are both constants, token values mapped to range values. for example LE in a range table is the same as <={LOW} as a token and vice versa. you can not change this otherwise the range tables are not valid anymore. recommend to add an extra description field to the output if you want to change the output for the user.

choper725:
ok ill remove the text elements, maybe will do that locally

PR #861, opened 2024-02-03
updates to driver js & imagemap CC

choper725:


choper725:
what should i change here ?? to fix the error

oblomov-dev:
thank you!

PR #911, opened 2024-02-19
New Layout

abapsheep:


oblomov-dev:
updated a few things, you can use some functions directly from z2ui5_cl_util. will check the other issues.

oblomov-dev:
finally it works now! I will probably replace the database tables with the table of the utility functions, that we have just one persistance for generic data in the project. but for now it looks good, thank you for the PR!

PR #926, opened 2024-02-21
New wizard control

axelmohnen:
New control: sap.m.wizard


oblomov-dev:
Thank you! 🙏
(needed to rename the helper function)

PR #943, opened 2024-02-23
refactoring

oblomov-dev:


larshp:
note https://github.com/abap2UI5/abap2UI5/pull/943/checks?check_run_id=21914369412, its likely that this PR contains a breaking change in the samples repo

oblomov-dev:
@larshp thanks for the hint, now also the cross check is green (maybe i will also add this check to the branch protection rule)
image

PR #964, opened 2024-02-27
update xml templating model binding & debugger feature for templating

choper725:
1. removed meta model name from preprocessors model in http_get. meaning no need to bind {meta>/} yet {/} is enough, once useing template preprocessor.

2. add logic to debbuger to identify once templating is being used and view xml source after templating
before templating:
image
after:
image

3. decode > in xml output of debugger.


oblomov-dev:
Thank you for all your work concerning Templating! It's good to have this concept integrated, it gives a lot more flexibility for creating views in the future. 

Yes nice idea, it's more straightforward like this and thank you also for extending the debugger tools. 🙏

PR #982, opened 2024-03-01
Update z2ui5_cl_xml_view.clas.abap

Marchl-Woid:
Change on Element proportion_zoom_strategy. Added property zoomLevel

Reason: Gantt Chart with Container ToolBar: Toolbar has the feature zoomlevel and can now be forwarded to the axis

image


oblomov-dev:
thank you! 🙏

PR #996, opened 2024-03-05
update http_get + add property to xml view

choper725:
1.
adjusted popup + popover fragment assign to sap.z2ui5.oViewPopup position in order to access view object before popup open (being able to run js on popup dom object before dialog open).
2. created a Fragment global var to get dom object as oViewPopup is not the actual view of the fragment, actually the oViewPopup is the dialog object from the popup.


oblomov-dev:
thank you!

PR #1000, opened 2024-03-08
Improve class F4 dialog - add class description

christianguenter2:
Before
image

After
image


larshp:
nice, some false positives, can you open an issue in abaplint repo?

christianguenter2:
I'll do tomorrow. But just adjusted the implementation.

oblomov-dev:
thank you, nice extension!

PR #1032, opened 2024-03-20
various updates

choper725:
1. fallback for popover openById in http_get claSS
2. update xml view class
3. corrected debugging cc logic upon errors.xml


oblomov-dev:
Thank you!

oblomov-dev:
image


nice one!

PR #1069, opened 2024-04-10
add properties to charts

jelliottp:
Add some missing properties to line chart, donut chart, and bar chart.


oblomov-dev:
Thanks!

PR #1070, opened 2024-04-10
spreadsheet columnconfig as a property of xml view

choper725:
add column config array to custom control as a binded table instead of local json string. (not breaking - kept the previous implementation version active as well)


oblomov-dev:
Thanks!

PR #1079, opened 2024-04-16
Use static types instead of magic constants

christianguenter2:


christianguenter2:
ah yes, we cannot reference the demo app. hm...

oblomov-dev:
lets switch to your snippet, its more beautiful. thanks.

PR #1081, opened 2024-04-16
catch cx_sy_rtti_syntax_error

christianguenter2:
If class has syntax error cx_sy_rtti_syntax_error is raised and propagated.
grafik

With this fix we catch it and deal with it as the class doesn't exist.


oblomov-dev:
Thanks

PR #1082, opened 2024-04-16
more fixes for renaming

christianguenter2:


oblomov-dev:
ah I am not sure if this works...
the method rtti_get_intfname_by_ref  returns 'OBJECT' because it imports an object reference

thats why I switched back to do it manually without a helper.

But I also could be wrong, did you tested for example the search app?

christianguenter2:
ah, yes. You're right. Tested something but obviously the wrong thing. To much renaming today :-)

christianguenter2:
`z3ui5_cl_util_api=>rtti_get_intfname_by_ref` can never work with `type ref to object` as a class can implement multiple interfaces?

PR #1095, opened 2024-04-21
update spreadsheet cc

choper725:


oblomov-dev:
thanks

choper725:
ohh yes you are right, ill try to check what an be done on cc load

choper725:
should now behave as expected with #1101

oblomov-dev:
Nice! works now on my end 👏
image

choper725:
i can try create a check module avaliable method as a custom cc, and we can probebly implement it in the xml view, to catch all of the non cc ciontrols who are not in openui5…
could be a better way tho..

for now can figure a way without try define the module with a try catch.

the part on whete and how to implement the check logic is the intresting one…

ill do something, we cam echance it from there,
let me know if you think of a better idea @oblomov-dev

PR #1156, opened 2024-06-04
various changes

choper725:
xml view
http get
client intf


oblomov-dev:
as always, thank you 🙏

PR #1172, opened 2024-06-07
two_way  binding based on param in core_types

choper725:
two_way param  binding based on param in core_types in http get method


oblomov-dev:
Thanks

PR #1176, opened 2024-06-10
New Layout

abapsheep:


oblomov-dev:
mhh did you pull latest updates? on my end it seems to work, but just tested it quickly:
image

yes we have two approaches now, @abapsheep just uses UI5 for the controls and does all the logic in abap by himself and going with the flexibiliy service would help to use the ui5 layout logic instead, but somehow the brain needs to be adapted 😅

choper725:
yes i pulled latest, ill put down exactly what expected and what i get using this.

acrtually using the standard controls for variant management & table column selection using generic abap brain, you dont need flexibility service for it to work.

PR #1190, opened 2024-06-13
downport to 752 compatible syntax

christianguenter2:
fix these errors
grafik


christianguenter2:
shouldn't this be discovered by the abaplint checks?

oblomov-dev:
thank you @christianguenter2 for the fixes!

and now i also see how helpful the whole time abaplint is, no chance to keep the project under control without it anymore :) thank you @larshp for this nice tool!

PR #1194, opened 2024-06-13
avoid /ui2/cl_json

oblomov-dev:


choper725:
i know i used it in the validator cc, its just a work in progress as i dont fully understand ajson and how to achive what i need the json to be from abap struc.

oblomov-dev:
@choper725  no problem, just do it how its the easiest way for you and we can try to refactor it later.

PR #1218, opened 2024-06-19
Update Layout and Transport Request App for Customizing

abapsheep:
A simple app with which you can select open customizing transport requests assigned to the user. The app also offers the option of saving changes to a table in transport requests. 
Unfortunately, the whole thing only works onPrem. I have not managed to get the cloud functions to work...


oblomov-dev:
thanks

PR #1291, opened 2024-07-25
Added property id to OverflowToolbarButton

cwildt:
Although the property id cannot be found in the official documentation for the OverflowToolbarButton, by adding it you can now also use functions that require an ID, such as the Action Sheet.


oblomov-dev:
thank you

PR #1297, opened 2024-07-30
Add new nethod & missing properties

francisco-milan:
-Add new method invisible text
-Add missing property to toolbar_spacer
-Add missing property to input


oblomov-dev:
thanks

PR #1306, opened 2024-08-10
Addition of the Side Navigation

LadwigDev:
Added elements to the view class:
Side Navigation
Navigation List
Navigation List Item
Fixed Item

Added parameter id to the flex_box method


oblomov-dev:
Thank you! 🙏

PR #1352, opened 2024-08-30
Data loss protection with dirty flag

### LINKED ISSUE #1348: FLP Integration - data loss protection
Do we have means in abap2UI5 (flp integration) to implement data loss protection on back navigation? Something like this?
grafik

If not, any pointers I could follow?

Edit: Seems like this does the job. Though I'm not sure how to integrate this in abap2UI5...
https://help.sap.com/doc/saphelp_nw74/7.4.16/en-US/5d/c265ddbafb424b961bdf8f7c6bbe15/content.htm?no_cache=true

christianguenter2:
This is a first try to implement kind of data loss protection, which triggers this popup when browser back or reload button is pressed, or FLP back navigation is used. 
image

Therefore I introduced a new method `z2ui5_if_client~dirty` and a new flag `dirty` which can be used to set the page dirty.
image

Note that we use `sap.ushell.Container.setDirtyFlag` while running in shell context, e.g. FLP, and `window.onbeforeunload` when running standalone. 

As I'm not too familiar with the inner workings of abap2UI5 I have know clue whether that's the right way to do it. Feel free to improve.

See also demo app z2ui5_cl_demo_app_279.

fixes #1348

### DIFF
--- a/src/01/02/z2ui5_cl_core_client.clas.abap
+++ b/src/01/02/z2ui5_cl_core_client.clas.abap
@@ -337,4 +337,9 @@ CLASS Z2UI5_CL_CORE_CLIENT IMPLEMENTATION.
 
   ENDMETHOD.
 
+
+  METHOD z2ui5_if_client~dirty.
+    mo_action->ms_next-s_set-dirty = val.
+  ENDMETHOD.
+
 ENDCLASS.
--- a/src/01/02/z2ui5_cl_core_http_get.clas.abap
+++ b/src/01/02/z2ui5_cl_core_http_get.clas.abap
@@ -398,6 +398,16 @@ CLASS Z2UI5_CL_CORE_HTTP_GET IMPLEMENTATION.
                `            if (sap.z2ui5.oResponse.PARAMS?.S_VIEW?.CHECK_DESTROY) {` && |\n| &&
                `                sap.z2ui5.oController.ViewDestroy();` && |\n| &&
                `            };` && |\n| &&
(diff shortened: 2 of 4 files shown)


christianguenter2:
See https://github.com/abap2UI5/abap2UI5-samples/pull/342

oblomov-dev:
Thank you for your work! the code works well and its a nice feature :)

I put the code slightly to another position and created a custom control for it. Then we do not need to extend the client interface with an additional method. You can access this feature now similar to the focus or title by binding it through the view. 

updated the repository and sample, check out the data loss branches.

if needed, you can now also use expression binding and for example do:
```
  page->_z2ui5( )->dirty(  '{= $' &&  client->_bind_edit( text_input ) && ' !== "" }' ).
```
that means after making an input, you don't need to do a server roundtrip before the data loss protection is active for the refresh button. 

Further improvements could be registrating on the model change event at the frontend and automatically activate the data loss protection if any editable field is changed without server roundtrip, but i think this is already good enough for most of the use cases. 

does this work for you or did i miss something? thank you for the PRs.

christianguenter2:
Amazing! Works as expected. I knew there ought to be a better way :-) Thx a lot

Added also regular binding to the sample.
`page->_z2ui5( )->dirty( client->_bind_edit( dirty ) ).`

oblomov-dev:
great! its merged now.
@christianguenter2 maybe i make a little sunday linkedin tweet about this feature. can i name you as the PR creator or should i make it anonym?

christianguenter2:
Sure. Go ahead :-) 
https://www.linkedin.com/in/christian-g%C3%BCnter-a86b0b100/

PR #1359, opened 2024-09-03
Progress Indicator Property Visibility

Marchl-Woid:
Addes Progress Indicator Property Visibility


oblomov-dev:
thank you for the PR! fixed the issue by renaming zqmcafw_cl_util to z2ui5_cl_util

PR #1427, opened 2024-09-13
Popup to select: New feature multiselect

christianguenter2:
New optional parameter `multiselect` of `z2ui5_cl_pop_to_select=>factory` to enable selection of multiple entries. In this case the field `table` of the result structure is filled instead of `row`.

image

sample z2ui5_cl_demo_app_152 update here
https://github.com/abap2UI5/abap2UI5-samples/pull/362


oblomov-dev:
nice! thank you

PR #1498, opened 2024-10-11
fix cloud

christianguenter2:
I tried to run the project in BTP ABAP and S/4 public cloud. Needed to make these adjustments to make it run.


oblomov-dev:
ah you were the first one testing the new handler in abap cloud, thanks

PR #1554, opened 2024-11-21
Update z2ui5_cl_xml_view.clas.abap

Marchl-Woid:
Visible Property to Label

PR #1555, opened 2024-11-21
Update z2ui5_cl_xml_view.clas.abap

Marchl-Woid:
Visible Property in Label


oblomov-dev:
see https://github.com/abap2UI5/abap2UI5/pull/1554#event-15383167824

PR #1563, opened 2024-11-21
enable toggle debug popup

christianguenter2:
Before this change it was possible to open multiple debug windows by pressing CTRL+F12 several times. 
After this change only one debug popup can be opened, pressing CTRL+F12 again closes the debug popup.


oblomov-dev:
thanks

PR #1596, opened 2024-11-30
Properties for Spot

Marchl-Woid:


oblomov-dev:
thank you for the PRs 🙏

PR #1630, opened 2024-12-10
SmartMultiInput Enablement

mzinnoecker:
- added SmartMultiInput to View
- added SmartMultiInput extension for rangeData/token handling similar to MultiInput
- added annotationURI, as we also need the annotations from the backend
- demo in Z2UI5_CL_DEMO_APP_319 (separate pull request in sample repo)


mzinnoecker:
Ohhh this is nice! Sorry for the extra work, did not read it in detail yet tbh :)

PR #1657, opened 2024-12-15
fix broken debug view close

christianguenter2:
Before this fix closing the debug popup with the close button was broken. Reopen required to press `CTRL+F12` twice.


oblomov-dev:
Thanks

PR #1782, opened 2025-02-22
Update App.controller.js

feng1976:
No display camera window in demo Z2UI5_CL_DEMO_APP_306. so modify App.controller.js of line 903/914


oblomov-dev:
thanks!

PR #1853, opened 2025-05-19
Fix open abap

oblomov-dev:


src/01/02/z2ui5_cl_core_client.clas.abap
```diff
@@ -44,7 +44,12 @@ CLASS z2ui5_cl_core_client IMPLEMENTATION.
       ).
 
     TRY.
+
+        IF mo_action->mo_http_post->ms_request-s_front-o_comp_data IS NOT BOUND.
+          RETURN.
+        ENDIF.
         DATA(lo_params) = mo_action->mo_http_post->ms_request-s_front-o_comp_data->slice( `/startupParameters/` ).
```
larshp, line 51: maybe something like the below is easier to maintain

```suggestion
       DATA(foo) = mo_action->mo_http_post->ms_request-s_front-o_comp_data.
        IF foo IS NOT BOUND.
          RETURN.
        ENDIF.
        DATA(lo_params) = foo->slice( `/startupParameters/` ).
```

oblomov-dev:
yes looks better like this!

PR #1866, opened 2025-06-07
add onAfterRoundtrip

choper725:
not sure about the position, just added it after Server.Roundtrip() call

### DIFF
--- a/src/01/03/z2ui5_cl_app_view1_js.clas.abap
+++ b/src/01/03/z2ui5_cl_app_view1_js.clas.abap
@@ -14,7 +14,9 @@ CLASS z2ui5_cl_app_view1_js DEFINITION
 ENDCLASS.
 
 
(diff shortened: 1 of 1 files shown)


oblomov-dev:
fyi, this is then the auto created PR which updates the stringified JS based on the app folder:
https://github.com/abap2UI5/abap2UI5/pull/1867

in this case it was just some formatting, because we also updated the abap strings.

choper725:
oh, so if i working on a pr in Z2UI5_CL_APP_APP_JS

i should pr in another place also ?

back at it need to learn how it works and whats changed in the past year :)

choper725:
oh ok, need to see how i do that,

for now i finished updating uitableext js cc in the app_app class
i will pr it and next time ill update from the frontend way you shared.

oblomov-dev:
👍

PR #1889, opened 2025-06-25
New method ROW_SETTINGS

axelmohnen:
The new method ROW_SETTINGS represents sap.ui.table.RowSettings. This can be used for row highlighting in sap.ui.table.


oblomov-dev:
thanks 🙏

PR #1913, opened 2025-07-08
fix dataloss protection for standalone apps using SAPUI5

christianguenter2:
Before this fix, data loss protection (sample `z2ui5_cl_demo_app_279`) didn't work when bootstrapped with SAPUI5 and running standalone. It's not sufficient to load `sap.ushell.Container`, because `Container.setDirtyFlag(val)` only works within FLP.


oblomov-dev:
thanks

Issue #2010, opened 2025-10-13
Using ImageEditor


Nioniotte:
Hello Everyone,

Objective: the user uses the camera to take a photo and then edits the image using the editor. After editing, save the image in SAP (GOS or ArchiveLink for example). 

For do that, I want to use the ImageEditor: https://sapui5.hana.ondemand.com/1.71.74/#/api/sap.suite.ui.commons.imageeditor.ImageEditor/overview in a 2UI5 app.

First, I want to test it and I think I need to call the editor like this:

``...)->page( title = 'abap2UI5 - test'``
``      )->_generic(``
``      EXPORTING``
``        name   = 'ImageEditor'``
``        ns     = 'commons'``
``        t_prop = VALUE #(``
``            ( n = `src`    v = base_url && `/test-resources/sap/m/demokit/sample/ImageContent/images/ProfileImage_LargeGenTile.png` )        )``
``        )->stringify( ) ).``

But in property name with value = 'ImageEditor' is see a not found in the console:
ui5loader-dbg.js:1338   GET https://xxx/sap/public/bc/ui5_ui5/resources/sap/suite/ui/commons/ImageEditor.js net::ERR_ABORTED 404 (NOT FOUND)

It is because the module are here: Module: sap/suite/ui/commons/imageeditor/ImageEditor

If I change the value of name to 'imageeditor/ImageEditor', I have this UI5 (or abap2UI5) error: The following problem occurred: XML parse Error for code: -1 reason: xmlns: URI sap.m is not absolute src: line: 1 linepos: 59 filepos: -1.

To be honest, I have no idea how to do it and whether it's possible.

Thanks for your help,
N

christianguenter2:
should be possible with `deviceId`

https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
<img width="1404" height="549" alt="Image" src="https://github.com/user-attachments/assets/6f21e2fb-4292-4422-bdb4-6466acdd40ae" />

https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices

christianguenter2:
> I see back or front. This is for mobile device right? But on computer?

this is mobile. Computer looks like this on my machine:

<img width="640" height="174" alt="Image" src="https://github.com/user-attachments/assets/bcfae7b2-4229-47ab-afc8-c56908cc6ef1" />

oblomov-dev:
[...]

PR #2021, opened 2025-10-17
downport

christianguenter2:
#2010 

works with new releases
image

but with 1.71.74 I get this error when opening image_editor popup:
image
Any idea?

### DIFF
--- a/src/02/01/z2ui5_cl_pop_image_editor.clas.abap
+++ b/src/02/01/z2ui5_cl_pop_image_editor.clas.abap
@@ -111,8 +111,8 @@ CLASS z2ui5_cl_pop_image_editor IMPLEMENTATION.
                                       contentwidth  = `80%` ).
 
     popup->image_editor_container(
-           enabledbuttons = mv_enabledbuttons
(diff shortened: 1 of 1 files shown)


christianguenter2:
same in others samples like `z2ui5_cl_demo_app_060`. Hm...

oblomov-dev:
mhh yes sending JS to the frontend is a good trick as an last option, but unfortunately it also creates problems here and there...

did you tried to use two-way data binding here?
https://github.com/christianguenter2/ABAP2UI5/blob/main/src/02/01/z2ui5_cl_pop_image_editor.clas.abap#L118

Maybe then the sendImage function is not needed anymore.

oblomov-dev:
> Yeah, tried with binding. Didn't work. SAP also uses getImagePngDataURL() in their sample.

ah too bad! another option is to create a custom control, i did this for example for the message manager, because i was also not able to bind some properties and needed to call methods of the control.

but i think the frontend action is also a good solution. i merge it, as always thanks a lot, nice work, big help 🙏

PR #2198, opened 2026-04-15
Update z2ui5_cl_pop_table.clas.abap

axelmohnen:
Use the DDIC field label instead of component names in column header


oblomov-dev:
thanks

PR #2401, opened 2026-07-11
update abaplint

oblomov-dev:
AI said:

# check_syntax: false positive "must be escaped in strict mode" on downported v702 SELECT

## Summary

Since **@abaplint/cli 2.119.30**, the `check_syntax` rule reports a false
positive on plain (non-strict) Open SQL when the configured syntax version is
**v702**. It demands `@` host-variable escaping in a `SELECT ... WHERE col IN
range` statement, even though inline `@` escaping is a 7.40+ feature and is not
valid in v702 at all. This makes it impossible to satisfy the check: adding `@`
produces a v702 syntax error, and omitting it produces the reported error.

## Affected versions

- **Last good:** 2.119.29
- **First bad:** 2.119.30
- Still present in 2.119.59 (latest at time of writing)

Bisected by running the same config/fix against each patch release.

## Where it surfaces

In the abap2UI5 project, the `auto_downport` step runs `abaplint --fix` with a
`v702` config. The `downport` rule rewrites a modern escaped SELECT into old
7.02 syntax, and `check_syntax` (in the same `--fix` pass) then rejects the
downported output.

Source (valid, modern, uses `@`):

```abap
METHOD load_multi_by_handle.
  DATA lr_uname   TYPE RANGE OF z2ui5_t_91-uname.
  DATA lr_handle  TYPE RANGE OF z2ui5_t_91-handle.
  DATA lr_handle2 TYPE RANGE OF z2ui5_t_91-handle2.
  DATA lr_handle3 TYPE RANGE OF z2ui5_t_91-handle3.
  " ... ranges filled ...
  SELECT FROM z2ui5_t_91
    FIELDS *
    WHERE uname   IN @lr_uname
      AND handle  IN @lr_handle
      AND handle2 IN @lr_handle2
      AND handle3 IN @lr_handle3
    INTO CORRESPONDING FIELDS OF TABLE @result.
ENDMETHOD.
```

After the `downport` rule rewrites it to v702 (correctly dropping the `@`,
which v702 does not support):

```abap
SELECT * FROM z2ui5_t_91 INTO CORRESPONDING FIELDS OF TABLE result
  WHERE uname   IN lr_uname
    AND handle  IN lr_handle
    AND handle2 IN lr_handle2
    AND handle3 IN lr_handle3.
```

`check_syntax` reports on this valid v702 code:

```
z2ui5_cl_util_db.clas.abap[164, 5] - SELECT: "lr_uname"   must be escaped with @ in strict mode (check_syntax) [E]
z2ui5_cl_util_db.clas.abap[164, 5] - SELECT: "lr_handle"  must be escaped with @ in strict mode (check_syntax) [E]
z2ui5_cl_util_db.clas.abap[164, 5] - SELECT: "lr_handle2" must be escaped with @ in strict mode (check_syntax) [E]
z2ui5_cl_util_db.clas.abap[164, 5] - SELECT: "lr_handle3" must be escaped with @ in strict mode (check_syntax) [E]
```

## Minimal reproduction

`abaplint.jsonc`:

```jsonc
{
  "syntax": { "version": "v702", "errorNamespace": "." },
  "rules": { "check_syntax": true }
}
```

`zcl_test.clas.abap` (a plain, non-strict v702 SELECT with a range):

```abap
CLASS zcl_test DEFINITION PUBLIC FINAL CREATE PUBLIC.
  PUBLIC SECTION.
    CLASS-METHODS run.
  PROTECTED SECTION.
  PRIVATE SECTION.
ENDCLASS.

CLASS zcl_test IMPLEMENTATION.
  METHOD run.
    DATA lr_carrid TYPE RANGE OF sflight-carrid.
    DATA lt_result TYPE STANDARD TABLE OF sflight.
    SELECT * FROM sflight INTO TABLE lt_result
      WHERE carrid IN lr_carrid.
  ENDMETHOD.
ENDCLASS.
```

- `npx @abaplint/cli@2.119.29` -> 0 issues
- `npx @abaplint/cli@2.119.30` (and later) -> "must be escaped with @ in strict mode"

## Expected behaviour

Under `syntax.version: v702`, a plain SELECT without any strict-mode syntax
element (no `@` escaping, no comma-separated field list, no inline `INTO @DATA`)
must **not** be flagged as strict, and host variables in `IN <range>` /
`WHERE col = var` must not be required to carry `@`. The "strict mode escaping"
requirement should only apply when the SELECT actually uses strict syntax, and
never for versions below v740 where `@` escaping is not permitted.

## Impact / workaround

Any codebase that lints downported v702 output (e.g. abap2UI5's `auto_downport`
pipeline) breaks on upgrade past 2.119.29. As a temporary workaround we pinned
`@abaplint/cli` to `>=2.119.24 <2.119.30`; we'd like to lift that cap once this
is resolved.


larshp:
to reproduce

```abap
DATA tab TYPE voided.
DATA result TYPE voided.
SELECT FROM sdfsdfsdf
  FIELDS *
  WHERE uname IN @tab
  INTO CORRESPONDING FIELDS OF TABLE @result.
```

PR #2489, opened 2026-07-29
Fix stale check_csrf_active wording and repair the red browser test suite

oblomov-dev:
# Description

Two independent fixes, neither changing product behaviour.

### 1. `check_csrf_active` comments (`858f3c3`)

`check_csrf_active` was opt-in when CSRF defense arrived in #2479. #2486 changed it to be seeded with `abap_true` in `z2ui5_cl_exit=>set_config_http_post` before the user exit runs, making it on-by-default with an opt-out. Two comments still described the old behaviour:

- `src/02/z2ui5_if_types.intf.abap` — called the field *"when set via the exit"* and *"Opt-in"*.
- `src/02/z2ui5_cl_http_handler.clas.testclasses.abap` — `test_csrf_inactive` was labelled *"opt-in"*.

The already-correct comments (`z2ui5_cl_http_handler.clas.abap:54`, `z2ui5_cl_exit.clas.abap:118`, `z2ui5_cl_exit.clas.testclasses.abap:60`) were left alone. `check_hide_error_details` still reads *"when set via the exit"*, which is correct for that field — it does default to `abap_false`.

### 2. `test_browser` (`7fb8294`)

`test_browser` has failed on **every one of the last 30 runs**, across all branches — 15 failures (5 tests × 3 browsers). All of them are stale assertions, not product defects.

**`error-view.spec.js` (4 tests).** #2469 made `ErrorView.show` prefer the friendly UI5 dialog, falling back to the raw-DOM overlay only when `sap.m` cannot be resolved. In a real browser those controls are always loaded, so the overlay these tests assert on was never built. The fix denies just the `sap/m/*` lookups for the duration of the `show()` call — the synchronous form returns `undefined` so `showFriendlyDialog` bails, the array form takes the errback so `loadFriendlyDialogAsync` bails — which is precisely the last-resort path the overlay exists for. `sap.ui.require` is restored immediately after, so UI5's own module loading is untouched. This preserves the split documented in `errorView.spec.js`, which states the raw overlay is covered by the browser suite.

**`roundtrip.spec.js` (1 test).** The case asserted the 500 body contains `"abap2UI5"`. The top-level catch in `z2ui5_cl_http_handler=>_main` answers with the raw exception text — for a malformed body that is the ajson message `Json parsing error: Not JSON @Line 1, Offset 1`. No product marker, and none was ever promised (the body was already `lx->get_text( )` before #2486). The assertion now targets the framework error text, keeping the point of the case: a broken body yields a 500 carrying the real reason instead of the generic SAP ICF error page.

## How to test

- The `roundtrip` case was reproduced failing on `origin/main` and passing with this change, run locally against the transpiled backend (`auto_downport` + `auto_transpile` + Playwright).
- The 317 JS unit tests still pass.
- The four `error-view` cases could **not** be exercised locally — they need UI5 from `sdk.openui5.org`, which the dev container cannot reach (proxy returns 403), so `beforeAll` times out there for an unrelated reason. They rest on the code path described above; CI is the check.

## Checklist

- [ ] `npx abaplint` passes (0 issues) — not run locally; the abaplint check on this PR is green
- [x] No manual edits under `src/00/` or `src/01/03/` (see AGENTS.md)
- [x] Public API in `src/02/` only changed additively — comment text only

### DIFF
--- a/node/tests/e2e/error-view.spec.js
+++ b/node/tests/e2e/error-view.spec.js
@@ -37,7 +37,39 @@ async function showErrorView(/** @type {boolean} */ withRetry) {
             const options = retry
               ? { onRetry: () => (window["__retried"] = true) }
               : undefined;
-            ErrorView.show("something broke", undefined, options);
+            // ErrorView.show prefers the friendly UI5 dialog and only builds
+            // the raw-DOM overlay tested here when sap.m cannot be resolved.
+            // In a real browser those controls are always loaded, so deny just
+            // the sap/m/* lookups for the duration of this one call: the
+            // synchronous form returns undefined (showFriendlyDialog bails)
+            // and the array form takes the errback (loadFriendlyDialogAsync
+            // bails), which is exactly the last-resort path the overlay exists
+            // for. Restored right after, so UI5's own module loading - which
+            // also goes through sap.ui.require - stays untouched.
+            const realRequire = window["sap"].ui.require;
+            const isControl = (/** @type {any} */ dep) =>
+              typeof dep === "string" && dep.startsWith("sap/m/");
+            window["sap"].ui.require = function (
+              /** @type {any} */ deps,
+              /** @type {any} */ cb,
+              /** @type {any} */ errback,
+            ) {
+              if (typeof deps === "string") {
+                return isControl(deps)
+                  ? undefined
+                  : realRequire.call(this, deps);
+              }
+              if (Array.isArray(deps) && deps.some(isControl)) {
+                if (errback) errback(new Error("sap.m denied by test"));
+                return undefined;
+              }
+              return realRequire.call(this, deps, cb, errback);
+            };
+            try {
+              ErrorView.show("something broke", undefined, options);
+            } finally {
+              window["sap"].ui.require = realRequire;
+            }
             resolve(undefined);
           },
         );
--- a/node/tests/e2e/roundtrip.spec.js
+++ b/node/tests/e2e/roundtrip.spec.js
@@ -60,9 +60,7 @@ test("chains the session draft across roundtrips", async ({ request }) => {
   expect(second.S_FRONT.ID).not.toBe(first.S_FRONT.ID);
 });
 
-test("returns the backend-built view XML on app start", async ({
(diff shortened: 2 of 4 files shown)


oblomov-dev:
`test_browser` is red here, but it fails identically on `main` and is not caused by this PR.

This PR's diff is comment text only — no executable line changed — so it cannot affect Playwright e2e behaviour.

**Same 15 failures, same 45 passed, on both:**

| | this PR ([job](https://github.com/abap2UI5/abap2UI5/actions/runs/30469038450/job/90634184016)) | `main` @ 5d2c67a ([job](https://github.com/abap2UI5/abap2UI5/actions/runs/30429021175)) |
|---|---|---|
| `error-view.spec.js:49` `:62` `:70` `:81` | failed ×3 browsers | failed ×3 browsers |
| `roundtrip.spec.js:116` | failed ×3 browsers | failed ×3 browsers |
| passed | 45 | 45 |

The `roundtrip.spec.js:116` assertion is `expect(await res.text()).toContain("abap2UI5")`, and the received body is `Json parsing error: Not JSON @Line 1, Offset 1` — a 500 whose body no longer carries the framework marker the test looks for.

Fixing that is out of scope for a comment-only change, so I am not touching it here. The other 11 checks are green (`abaplint / cross check` is `neutral`, not a failure).

---

oblomov-dev:
Correction to my comment above, on two points.

**The cited comparison run was the wrong reference.** I pointed at run `30429021175` on `5d2c67a` as "`main`". That commit is not an ancestor of `main` — it descends from `b1ad4c3` (#2483), while this PR's base `3db417c` came via #2487/#2488. GitHub reported the run's `head_branch` as `main`, which I took at face value.

The conclusion held up under a proper check: `roundtrip.spec.js:116` was reproduced failing on the actual base `3db417c`, run locally against the transpiled backend, and all 30 preceding `test_browser` runs were red across every branch.

**"Fixing that is out of scope" no longer describes this PR.** It was repointed to fix the suite: `7fb8294` repairs all 5 stale cases, and `test_browser` passed on the merged head.

---

PR #2581, opened 2026-08-14
Ask before initialising the Std media range set

oblomov-dev:
# Description

`createDeviceModel` calls `Device.media.initRangeSet("Std")` unconditionally. `"Std"` is UI5's own predefined range set, so by the time the device model is built something has usually initialised it already — UI5 then logs

```
Range set Std has already been initialized - DEVICE.MEDIA
```

and returns (`sap/ui/Device.js`). One info line in the console of every app start, for a call that does nothing.

Guarded with `Device.media.hasRangeSet`, which exists well before the manifest's `minUI5Version` of 1.71 (checked against `@openui5/sap.ui.core@1.71.0`). The call still runs on a runtime that has not set the range set up yet, so `{device>/media/range}` bindings behave the same either way.

`src/01/03/z2ui5_cl_ui5f_models_js.clas.abap` is the regenerated embedded copy — `npm run app2abap`, no manual edit.

## How to test

`node/tests/deviceModel.spec.js` covers both branches: the range set is still initialised when it is missing, and left alone when it is already there. The spec's `Device` stub grew a `hasRangeSet`.

Reproducing the message itself needs a browser: open any app and look for `Range set Std has already been initialized` in the console. I checked it against a generated BSP tree served page for page, with UI5 from a local OpenUI5 tree — the line is there before this change and gone after, and `/media/range` is still seeded before the first render.

## Checklist

- [x] `npm run verify:full` passes — *running at the time of writing; the gates that had completed were green (abaplint, naming, dynamic names, icons, abapgit format, ATC, frozen paths, guide, formatter scope, api snapshot, ABAP standard, ABAP cloud). I will report the final result in a comment.*
- [x] Unit tests added/updated where it makes sense (`node/tests/deviceModel.spec.js`, 766 specs green)
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` — regenerated via `npm run app2abap`

https://claude.ai/code/session_01Mwc1AhTXb3vmV5i8iVWL1v

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mwc1AhTXb3vmV5i8iVWL1v)_

### DIFF
--- a/app/webapp/model/models.js
+++ b/app/webapp/model/models.js
@@ -23,7 +23,14 @@ sap.ui.define(
         const oModel = new JSONModel(Device);
         oModel.setDefaultBindingMode("OneWay");
 
-        Device.media.initRangeSet(RANGE_SET);
+        // "Std" is predefined, so whoever touched Device.media first has
+        // usually initialised it already - and initialising it twice makes
+        // UI5 log "Range set Std has already been initialized" into every
+        // app start. Ask first; the call is still needed for the runtime
+        // that has not set it up yet.
+        if (!Device.media.hasRangeSet(RANGE_SET)) {
+          Device.media.initRangeSet(RANGE_SET);
+        }
 
         const refresh = () => {
           // Device.media exposes methods only - there is no property a binding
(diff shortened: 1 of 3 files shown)


oblomov-dev:
`npm run verify:full` finished green (exit 0), so the first checklist item stands without the caveat.

---

Issue #56, opened 2023-03-02
Switch Binding


sun-tassein:
I am not able to bind the switch value with the backend variable value.
I dont think it is working in the z2ui5_cl_app_demo_02 example either.

oblomov-dev:
thanks for the hint, it's a bug! 
i fixed it and you can try it again. 
but don't forget to change the binding mode to this:
<img width="954" alt="image" src="https://user-images.githubusercontent.com/102328295/222652927-4dc2d8b8-aab1-4fc0-9858-a2c607a1f951.png">

sun-tassein:
Thank you, it works now.

PR #2654, opened 2026-08-22
MultiInputExt: the suggestion-row half of addValidator, and the app rules the new ports need

oblomov-dev:
# Description

Five changes that came out of porting batch b51 in `samples-controls`, where the
framework side of the work lives.

**`app/webapp/cc/MultiInputExt.js` — cover the suggestion-row branch of `addValidator`.**
The control installed exactly one validator, and it only handled the free-text
branch: `args.text` became a `Token` with key and text both set to the typed
string. `sap.m.MultiInput` calls the same validator a second way — with
`args.suggestionObject`, the row the user picked out of a tabular suggestion
popup — and that branch was ignored, so picking a row produced a token carrying
the raw column text or nothing usable at all.

Two optional properties say which cells of a picked row become the token:

- `TokenKeyCell` (int, default `-1`) — the cell index the token key comes from
- `TokenTextCells` (string, default `""`) — the cell indices the token text is
  built from

Both default to the old behaviour, so an app that sets neither is unchanged.
`tokenFromRow(row)` reads them; the free-text branch is untouched, and app 040
in `samples-controls` is its regression test.

**`.github/abaplint/app-rules.json` — the shared `7bit_ascii` exclusion follows
the new ports.** Eleven more sample ports carry non-ASCII characters that come
straight from the original demo-kit mock data (`9.600 × 9.600 dpi`, en dashes in
UI5's own bootstrap prose, `…` in a navigation label), the same reason the
twenty-one names already on the list are there. This repository is the source of
that block; see "How to test" for what that means for the gate.

**Three backlog items and one `.gitignore` line** — the tabular-suggestion
validator itself, the linter's `MultiInputExt` mirror having to follow the two
new properties (now done, and the item goes with it), `DELETE … INDEX sy-tabix`
inside a `LOOP` written up a fourth time, and the three artefacts
`samples-controls`' e2e build writes into this tree.

## How to test

- `npm run check:js` — 892 frontend tests, including `MultiInputExt`'s.
- `npm run check:frontend` and `npm run check:app2abap` leave the tree clean.
- Behaviour: an app that binds a `MultiInputExt` to a tabular suggestion list and
  sets `TokenKeyCell`/`TokenTextCells` now gets a token built from the picked
  row; with neither property set, the free-text path behaves exactly as before.

`npm run gates` reports two failures worth naming up front:

- **`check:shared`** goes red on the `app-rules.json` block until `samples` and
  `samples-stack` copy it. That is by design for a `section` entry — the gate's
  own comment calls it "a gate-and-fix-by-hand affair", and this repository being
  the source means the source moves first. `samples-controls` already carries the
  new value on its open pull request; the other two consumers need the same
  one-line copy, and I do not have access to those repositories from here.
- **`check:counts`** is green in CI today and goes red once
  `samples-controls`' pull request lands: `llms.txt`,
  `docs/agents/building-apps.md` and `.claude/skills/build-an-app/SKILL.md` all
  say 416 ports, and that corpus is at 622. The gate reads the number from
  `samples-controls@main`, so bumping it here before that merge would break the
  gate rather than fix it. It is a follow-up, not part of this change.

## Checklist

- [x] `npm run verify` passes (`npm run verify:full` when `app/webapp/` changed) — apart from the two cross-repository gates described above
- [x] Public API in `src/02/` only changed additively — untouched here

---
_Generated by [Claude Code](https://claude.ai/code/session_01RRPQ84UVzuc2HqRLPt1T1V)_


oblomov-dev:
`check_gates` is red on two gates. One is this change's, one is not — and I got the second one wrong in the description above, so here is the corrected reading.

**`check:shared` — 3 problems, and this change caused them.** The `app-rules.json` block now differs from all three consumers (`samples`, `samples-controls`, `samples-stack`). That is the expected shape for a `section` entry: this repository is the source, so it moves first and the copies follow by hand. `samples-controls` already carries the new value on abap2UI5/samples-controls#143 and clears the moment that merges; `samples` and `samples-stack` need the same one-line copy and nothing else. I have not opened those two pull requests — that is a decision for the maintainer, not something to do unasked in two more repositories.

**`check:counts` — 4 problems, and this change did not cause them.** I wrote above that this gate is green today and goes red after #143 lands. That was wrong: it is **already red on `main`**. `samples-controls@main` says 421 ports, `llms.txt`, `docs/agents/building-apps.md` (twice) and `.claude/skills/build-an-app/SKILL.md` all still say 416, and this pull request touches none of those files. Verified against both default branches rather than inferred.

I am deliberately **not** bumping 416 → 421 here. #143 takes that corpus to 622, so a bump to 421 would be stale within hours and would widen this pull request for a number it does not own. The right moment is one follow-up after #143 merges, straight to the final figure.

Everything else in the job is green: `check:frozen`, `check:abapgit`, `check:atc`, `check:naming`, `check:dynamic`, `check:prose`, `check:skills`, `check:conventions`, `check:scripts`, `check:samples-md`, `check:changelog`, `check:backlog`, `check:version`, `check:abap2ui5`, `check:api`, `check:guide`, `check:icons`, `check:modules`, `check:formatter`, `check:asserts`.

One observation from the log worth recording: `check:mirrors` reported

> the installed @abap2ui5/linter does not ship scripts/check-upstream.mjs (published from linter 0.2.3 on) — SKIPPED, not verified.

so the mirror gate is not actually verifying anything on this run. That resolves itself with the linter release the `MultiInputExt` change is queued behind.

---

oblomov-dev:
The two missing copies are now open, so `check:shared` has somewhere to converge:

- abap2UI5/samples#800
- abap2UI5/samples-stack#56
- abap2UI5/samples-controls#143 already carries the value

All three are the identical one-line change, verified with `node scripts/check-app-rules.mjs` in each checkout against an `abap2UI5` tree carrying this pull request — both report "this repository matches the shared rule set - OK" over all 187 rules. Both configs still parse as JSONC and the extended regex compiles.

Ordering, spelled out because every one of these gates reads the *other* side's `main`:

1. **#2654 merges** → `check:app-rules` goes green on #143, #800 and #56 with no further push.
2. Those three merge → `check:shared` goes green here.

Until step 1, #800 and #56 are red on their own `check:app-rules`, which is the same circularity from the other end and is called out in both descriptions. `check:counts` here stays red either way — it is red on `main` today, for a number this pull request does not touch.

---

PR #2655, opened 2026-08-22
counts: the corpus claims follow samples-controls to 622 ports

oblomov-dev:
# Description

`check:counts` has been red on `main`, and this fixes it. Four claims about
`samples-controls`' corpus size said **416**; that corpus has now landed its
remaining batches and its generated `STATUS.md` reports **622**.

- `llms.txt`
- `docs/agents/building-apps.md` (twice)
- `.claude/skills/build-an-app/SKILL.md`

The gate had been drifting for a while — it was already reporting 421 before
`abap2UI5/samples-controls#143` merged, so this goes straight to the current
figure rather than to the one the prose was behind by. The number is read from
`samples-controls`' own generated state block, which is where
`corpus-count-gate.mjs` reads it too; nothing is counted here.

## How to test

- `npm run check:counts` — was 4 problems, now "every corpus count matches the corpus - OK".
- `npm run check:prose` — unchanged, green.
- `npm run check:shared` — green here against an `app-template` checkout carrying its regenerated mirror; see the ordering note below.

`docs/agents/building-apps.md` is mirrored into `app-template`'s `AGENTS.md`,
which is **generated** there by `npm run agents` from this guide plus the shared
deviation list. That regeneration is open as **abap2UI5/app-template#17** and is
the one-line counterpart to this change. Until it merges, `check:shared` reports
the gap from this side and `check:agents` reports it from the other — the same
source-moves-first ordering as the `app-rules` block in #2654.

## Checklist

- [x] `npm run verify` passes (`npm run verify:full` when `app/webapp/` changed) — `app/webapp/` untouched; the gates this diff can move are listed above
- [x] Unit tests added/updated where it makes sense (`.testclasses.abap` / `node/tests/`) — prose-only change, no test surface


oblomov-dev:
`check_gates` is red, and it is red on exactly the one line this pull request predicted — nothing else moved.

**What this change fixed.** `check:counts` now reports *"6 claim(s) declared, 6 checked against 2 of 2 corpus/corpora — every corpus count matches the corpus - OK"*. That gate was failing on `main` with 4 problems before this.

**What is left, and why no push fixes it here.** `check:shared`, one problem:

```
  first difference at line 555
    here:         "…/abap2UI5/samples), 622 gate-verified"
    app-template: "…/abap2UI5/samples), 416 gate-verified"
```

and from the other side, on abap2UI5/app-template#17:

```
generate-agents: the mirrored half of AGENTS.md is out of date.
  first difference at AGENTS.md line 662
    here:      "…622 gate-verified"
    the guide: "…416 gate-verified"
  The guide is abap2UI5's docs/agents/building-apps.md (read from abap2UI5/main).
```

The same sentence, each gate reading the other repository's `main`. Merging this one resolves both: #17 already carries the regenerated mirror (produced by `npm run agents`, not transcribed), so nothing needs pushing on either branch.

Everything else in the job is green — `check:frozen`, `check:abapgit`, `check:atc`, `check:naming`, `check:dynamic`, `check:prose`, `check:skills`, `check:conventions`, `check:scripts`, `check:counts`, `check:samples-md`, `check:changelog`, `check:backlog`, `check:version`, `check:abap2ui5`, `check:api`, `check:guide`, `check:icons`, `check:modules`, `check:formatter`, `check:asserts` — and on #17, `check:pin` and `check:template` are green too.

`check:mirrors` still reports itself SKIPPED (*"the installed @abap2ui5/linter does not ship scripts/check-upstream.mjs"*), unchanged from #2654 and unrelated to this diff; it clears when the pinned linter moves to 0.3.0.

---

PR #2656, opened 2026-08-23
The user-exit tests own the statics they measure, and abap-check names the bundle it cannot see

oblomov-dev:
# Description

Both halves come from the same system run a user reported: 1084 methods, 1083 passed, one failed — plus eight SLIN warnings on `z2ui5_cl_abap2ui5_local`.

## `test_superseded_intf` failed on a system and passed in CI

```
Critical Assertion Error: 'Test_Superseded_Intf: ASSERT_EQUALS'
  Expected [sap_belize] Actual [sap_horizon]
  LTCL_TEST_USER_EXIT->TEST_SUPERSEDED_INTF, line 118
```

The asymmetry is the test's own, not the framework's. The dispatch under test is

```abap
IF gi_user_exit IS BOUND.          " the current interface
ELSEIF gi_user_exit_dep IS BOUND.  " the superseded one
```

and `get_instance( )` binds `gi_user_exit` to whatever exit class it **finds on the system**. The test placed its double in `gi_user_exit_dep` but never cleared `gi_user_exit`, so on a system carrying a customer exit the `IF` branch wins, the double is never reached, and the theme stays at the `sap_horizon` the shipped exit seeds — which the test then reads as *"the old interface was skipped"*. In CI no such class exists, so the `ELSEIF` runs and it is green.

The other three tests have the same exposure: they assert the **shipped defaults**, and an installed exit that touches the theme or the expiry would move them too. It happens that the one on the reporting system does not.

So `setup` parks the installed exit and clears both statics, and `teardown` puts it back — they are class-wide, so leaving them changed would leak into whatever runs next.

**Nothing in the production dispatch changes.** Preferring the current interface is the documented behaviour, and the superseded one is still the fallback.

## `abap-check` gains the artefact it does not cover

The same run reported eight SLIN warnings against `z2ui5_cl_abap2ui5_local` — POSIX regex twice, ABAP Doc position, `<CLASS>` not supported / not closed three times, a redundant `CONV string( )`. All four kinds are already in the catalogue, three of them under **Gated**, which is the right question to ask: why did `check:atc` not catch them?

Because it never sees that code in that shape, and the entry now says so.

- Over `src/` the gate holds. Every `FIND`/`REPLACE … REGEX` here carries `##REGEX_POSIX`, including the five in the vendored AJSON where the pragma sits on the statement's **continuation** line. A line-based grep reports those five as unannotated and is wrong — the gate flattens statements first. Measured before writing it down, because *"the mirror is missing pragmas"* was the obvious wrong answer.
- `abap2UI5-local` is a different artefact. `trigger_local.yaml` copies these sources there, and a script folds them into **one** class whose Local Implementations hold everything. The warnings carry character **offsets** into that folded include (`@16232`, `@38704`), not lines in any file here. The annotations that silence these in `src/` therefore do not survive or do not apply after the fold, and that is where the fix belongs.
- The `<CLASS>` pair is the one worth opening first: `<p class="shorttext synchronized">` is the standard ADT shorttext form, and a tag scanner splitting on whitespace reads the attribute name as a second tag — which would make it an artefact of the fold rather than of the doc comment.

The entry also records what was **not** verified: `abap2UI5-local` was not attached to the session, so the bundler itself was not read.

## How to test

- `npx abaplint abaplint.jsonc` — 0 issues over 258 files.
- `npm run check:atc` — 102 files, OK.
- `npm run check:skills` — every rule id the skills name exists.
- On a system with a customer exit installed: all four `ltcl_test_user_exit` methods pass, where `test_superseded_intf` failed before.

**`npm run unit` has not been run for this branch.** It reads `node/output`, and that directory was being rewritten by a corpus build throughout — running `auto_transpile` alongside it would have collided. The change is confined to a test class's setup/teardown and a skill document; it is still a gap in the evidence and is called out rather than glossed over.

## Checklist

- [x] `npm run verify` passes (`npm run verify:full` when `app/webapp/` changed) — `app/webapp/` untouched; see the note on `npm run unit` above
- [x] Unit tests added/updated where it makes sense (`.testclasses.abap` / `node/tests/`) — this change *is* the test fix

---
_Generated by [Claude Code](https://claude.ai/code/session_01RRPQ84UVzuc2HqRLPt1T1V)_

### DIFF
--- a/.claude/skills/abap-check/SKILL.md
+++ b/.claude/skills/abap-check/SKILL.md
@@ -450,6 +450,32 @@ pitfalls".
 - **ABAP Doc is parsed as HTML.** A literal `<`, `>` or `&` must be escaped as
   `&lt;`, `&gt;`, `&amp;`.
 
+**Not gated: the BUNDLE.** `check:atc` walks this repository's `src/` — that is
+its stated scope, and over `src/` it holds: every `FIND`/`REPLACE … REGEX` here
+carries `##REGEX_POSIX`, including the five in the vendored AJSON where the
+pragma sits on the statement's CONTINUATION line (a line-based grep reports
+those as missing; the gate flattens statements first, so it does not).
+
+`abap2UI5-local` is a different artefact. `trigger_local.yaml` copies these
+sources into that repository, where a script folds them into ONE class whose
+"Local Implementations" hold everything. A user's system run on
+`z2ui5_cl_abap2ui5_local` (2026-08-23) reported eight SLIN warnings against
+that folded include — POSIX regex twice, ABAP Doc position, `<CLASS>` not
+supported / not closed three times, a redundant `CONV string( )` — at character
+OFFSETS (`@16232`, `@38704`), not at lines in any file here. All four kinds are
+the families above, and all four are silenced in `src/` by a pragma, a
+`##REGEX_POSIX`, or an escape. **They come back after the fold**, so what the
+fold does to those annotations is the thing to establish; the sources on this
+side are clean and no change here would move them. The `<CLASS>` pair is worth
+looking at first: `<p class="shorttext synchronized">` is the standard ADT
+shorttext form, and a tag scanner that splits on whitespace reads the attribute
+name as a second tag — which would make it an artefact of the fold rather than
+of the doc comment.
+
+Not verified from here: `abap2UI5-local` was not attached to the session that
+wrote this, so the bundler itself was not read. What IS measured is that
+`src/` carries the annotations and that `check:atc` is green over it.
+
 ## 4. Downport and transpile — one source, three targets plus a JS runtime
 
 **Gate: this repo** — `npm run verify` builds all three targets and runs the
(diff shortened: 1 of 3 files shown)


oblomov-dev:
`check_gates` was red with two problems. One was mine and is fixed in `a5388a7`; the other is not this branch's and needs a merge elsewhere.

**`check:prose` — mine, and the gate was right.**

```
Prose naming an object this repository does not ship:
  .claude/skills/abap-check/SKILL.md:462  z2ui5_cl_abap2ui5_local
```

That is the entry's own point: `z2ui5_cl_abap2ui5_local` is the single class `abap2UI5-local` folds these sources into, and the passage is about findings that appear *there* while being annotated away *here*. The gate offers exactly the right escape — *"if it genuinely belongs to another repository, add it to EXTERNAL with the reason"* — so it is now on `EXTERNAL` with its reason. `check:prose` reports 18 files, every name resolves, 10 external names allowed.

**`check:shared` — red on `main` too, and this branch does not touch it.**

```
  first difference at line 555
    here:         "…/abap2UI5/samples), 622 gate-verified"
    app-template: "…/abap2UI5/samples), 416 gate-verified"
```

That is the counterpart of #2655, which merged here: the guide moved to 622 and **abap2UI5/app-template#17** carries the regenerated mirror but has not been merged. Verified against both default branches rather than inferred — `abap2UI5@main` line 570 says 622, `app-template@main` line 662 still says 416. Nothing on this branch touches that file, and merging #17 clears it.

Worth recording because it nearly fooled me: running `npm run check:shared` locally reports **green**, because the gate prefers a sibling checkout over github and the `app-template` checkout beside this one sits on the branch that already carries 622. Locally green is not CI green for any of these mirror gates — the same trap the `app-rules` block set in #2654.

`check:mirrors` still reports itself SKIPPED (*"the installed @abap2ui5/linter does not ship scripts/check-upstream.mjs"*), unchanged and unrelated to this diff.

---

PR #2687, opened 2026-08-30
Restore the two cross-repository checks #2685 left broken on main

oblomov-dev:
# Description

[#2685](https://github.com/abap2UI5/abap2UI5/pull/2685) merged 35 seconds after it was opened and six seconds after `check_gates` went red, so two failures it introduced reached `main`. Both are reproduced against `origin/main` before this change and green after it.

**1. `check:shared` was red — and stays red for every pull request until fixed.**

`docs/agents/building-apps.md` is mirrored into `abap2UI5/app-template`'s `AGENTS.md`, which **generates** its copy with `npm run agents`. Nothing in this repository can make the two agree, so the guide edit goes back out:

```
  this repository is the source — copy it there, or change it here first
```

The text is not lost. The four methods it documented — `_bind_path( )`, `client->get( )` (incl. `t_model_skipped`), `set_app_state_active( )` and `set_session_stateful( )` — are recorded in `check-guide-api`'s `UNDOCUMENTED` map, each marked `PENDING` with what it is. The reverse-direction check added in #2685 asks for them again the moment those entries are dropped. **The guide edit and the app-template regeneration are one change, to be made when both repositories can be touched.**

**2. `check:mirrors` was worse, because it did not fail.** It printed:

```
check-upstream: could not find GLOBAL_TARGETS in the frontendaction class
  — the embedding changed, update parseGlobalTargets
linter-mirror: the sources could not be read — SKIPPED, not verified.
```

`@abap2ui5/linter` finds the three whitelists it hand-mirrors by the **exact source text** `const GLOBAL_TARGETS = {` / `const CONTROL_METHODS = {` / `const BINDING_METHODS = {` in the embedded carrier (its `scripts/check-upstream.mjs`). Wrapping each literal in `Object.assign(Object.create(null), …)` broke all three lookups at once, and the gate degraded to *not checking* rather than to failing — a cross-repository check that stops checking without saying so.

The tables stay prototype-less, which was the point of that change: they are plain literals again, with `Object.setPrototypeOf(x, null)` directly after each. Same effect on the lookup, marker intact. The mirror check verifies again and reports all mirrors in sync.

## How to test

```sh
git checkout main && npm run gates      # check:shared fails; check:mirrors prints "SKIPPED, not verified"
git checkout claude/repository-analysis-improvements-cnhod3
npm run gates                            # 30 checked - all OK
npm run check:mirrors                    # "the linter's copies still match this tree - OK"
npm run verify                           # everything green
```

`npm run verify` was run in full on this branch and reports `verify: everything green`.

## Checklist

- [x] Unit tests added/updated where it makes sense (`.testclasses.abap` / `node/tests/`) — no behaviour change here; the 1&nbsp;001 JS specs and the transpiled suite are unchanged and green
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` (see AGENTS.md) — `src/01/03/z2ui5_cl_ui5f_ctrlcall_js.clas.abap` is regenerated with `npm run app2abap`
- [x] `changelog.txt` has a line under `unreleased` when this changes behaviour — no user-visible behaviour change; the frontend tables behave exactly as #2685 made them

---
_Generated by [Claude Code](https://claude.ai/code/session_01GmqgFmjaEtpEqoG1kooFuG)_

### DIFF
--- a/.github/scripts/check-guide-api.mjs
+++ b/.github/scripts/check-guide-api.mjs
@@ -74,6 +74,18 @@ const UNDOCUMENTED = new Map([
   ["view_destroy", "as nest_view_display - the slot form of popup_destroy( )"],
   ["popover_destroy", "as nest_view_display"],
   ["set_push_state", "browser-history detail of the routing block in section 7"],
+  // These four SHOULD be in the guide, and the text for them is written. It is
+  // not here yet because this file is mirrored into abap2UI5/app-template's
+  // AGENTS.md, which generates its copy with `npm run agents` - so editing it
+  // fails `npm run check:shared` until that regeneration lands, which is what
+  // it did on main between #2685 and this change. The guide edit and the
+  // app-template regeneration are one change, made when both repositories can
+  // be touched. Drop these four entries then; the check below asks for them
+  // again the moment somebody does.
+  ["_bind_path", "PENDING the app-template mirror sync - named form of _bind( path = abap_true )"],
+  ["get", "PENDING the app-template mirror sync - the request context, incl. t_model_skipped"],
+  ["set_app_state_active", "PENDING the app-template mirror sync - the bookmarkable app-state hash"],
+  ["set_session_stateful", "PENDING the app-template mirror sync - the sticky-session switch"],
   ["get_app", "reaches another app instance on the stack - get_app_prev( ) is the documented case"],
   ["check_app_prev_stack", "a stack predicate for get_app_prev( ), documented through it"],
 ]);
--- a/app/webapp/core/actions/ControlCall.js
+++ b/app/webapp/core/actions/ControlCall.js
@@ -145,19 +145,19 @@ sap.ui.define(
     // control method -> kinds of its positional args. Args beyond the
     // declared kinds are dropped; trailing args the caller did not send are
     // not passed at all (so `open()` stays a true no-arg call).
-    // Object.create(null), not a plain literal, and the same for the two
-    // whitelists below and for the handler map in FrontendAction.js.
-    // These tables are indexed with a name that comes off the wire, and a
-    // plain object answers for every key Object.prototype has: with a method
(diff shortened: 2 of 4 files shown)


oblomov-dev:
`check_gates` failed on `d0a4cb6`, and the cause was not in this branch — it was **red on `main` itself**. Pushed a fix; the PR now also carries it.

**What failed:** `check:skills`, on five rule ids:

```
class-constructor-visibility · value-header-default-reassigned
into-corresponding-inline-decl · validating-setter-out-of-range
absent-boolean-overrides-default
    … is not a rule of the pinned abap2ui5lint
```

**Why it is not this PR's:** all 32 steps of `check_gates.yaml` pass locally on `d0a4cb6`. CI checks the *merge* commit, and while this PR was open [#2686](https://github.com/abap2UI5/abap2UI5/pull/2686) merged into `main`. Running the gate on `origin/main` alone — nothing of this branch in it — exits 1 with the same five names, so every pull request opened against `main` fails until this is fixed.

**Root cause:** #2686 documents seven rules the linter shipped on 2026-08-30. npm `latest` for `@abap2ui5/linter` is **0.5.1, published 2026-08-27**, and `package.json` pins `^0.5.1` — so those names are true statements about the linter and still unknown to the pinned copy.

**The fix** is the gate's own `PENDING` map, which cleans itself up: it *fails* on an entry whose rule the pinned linter has, so `bump-linter.yaml` removes these five on the first release past 0.5.1. Its comment said “rules that exist in a **released** linter this repository has not pinned yet”; these are merged on the linter's main and not published at all, so the wording now says “upstream” and each entry records which case it is — widening the meaning is written down rather than assumed.

Carried here rather than split off because it is the same failure this PR already fixes twice: a cross-repository check that went red on `main` because a sibling repository moved and this one did not follow. `main` is merged into the branch as well, so CI now checks the tree `npm run verify` was run against — result: `verify: everything green`, 30/30 gates.

---

PR #2688, opened 2026-08-30
List the five 2026-08-30 linter rules the pinned version does not have

oblomov-dev:
# Description

**`main` is currently red, and every pull request opened against it fails.** `npm run gates` on `997e73d` reports `1 of 30 failed → check:skills`:

```
class-constructor-visibility · value-header-default-reassigned
into-corresponding-inline-decl · validating-setter-out-of-range
absent-boolean-overrides-default
    … is not a rule of the pinned abap2ui5lint
```

This is not a defect in any of those skills. [#2686](https://github.com/abap2UI5/abap2UI5/pull/2686) documents seven rules the linter shipped on **2026-08-30**, but npm `latest` for `@abap2ui5/linter` is **0.5.1, published 2026-08-27**, and `package.json` pins `^0.5.1`. Five of those names are therefore true statements about the linter and still unknown to the pinned copy.

(This fix was part of [#2687](https://github.com/abap2UI5/abap2UI5/pull/2687) but did not make it into the squash — `main` carries that PR's `ControlCall.js` change and not this one, so the gate is still failing.)

## The fix

`PENDING` in `skill-rule-gate.mjs` is the mechanism for exactly this, and it cleans itself up: the gate **fails** on an entry whose rule the pinned linter has, so `bump-linter.yaml` takes these five out on the first release past 0.5.1.

Its comment said “rules that exist in a **released** linter this repository has not pinned yet”. These are not released at all — they are merged on the linter's `main`. The wording now says “upstream” and each entry records which of the two cases it is, so widening the meaning is written down rather than assumed.

No skill text changes: the `Linter:` lines #2686 added are correct and stay as they are.

## How to test

```sh
git checkout main && npm run gates     # 1 of 30 failed → check:skills
git checkout claude/repository-analysis-improvements-cnhod3
npm run gates                           # 30 checked - all OK
npm run check:skills                    # "5 pending a pin bump" … "every rule id a skill names exists - OK"
```

`npm run verify` was run in full on this branch: `verify: everything green`.

## Checklist

- [x] Unit tests added/updated where it makes sense (`.testclasses.abap` / `node/tests/`) — n/a, one gate exemption list
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` (see AGENTS.md) — none, `src/` untouched
- [x] `changelog.txt` has a line under `unreleased` when this changes behaviour — no user-visible behaviour change

---
_Generated by [Claude Code](https://claude.ai/code/session_01GmqgFmjaEtpEqoG1kooFuG)_

### DIFF
--- a/.github/scripts/skill-rule-gate.mjs
+++ b/.github/scripts/skill-rule-gate.mjs
@@ -56,14 +56,33 @@ const NOT_A_RULE = new Map([
   ['unit-tests', 'the abapGit round-trip rule name in this skill\'s own table'],
 ]);
 
-/* Rules that exist in a released linter this repository has NOT pinned yet.
- * Same semantics as a baseline entry: once the pin catches up the rule is
- * known, and an entry left behind here FAILS rather than lingering. */
+/* Rules that exist upstream but not in the linter this repository has pinned -
+ * either released under a version not pinned yet, or merged on the linter's
+ * main and not published at all. Same semantics either way: once the pin
+ * catches up the rule is known, and an entry left behind here FAILS rather
+ * than lingering.
+ *
+ * The second case is why the wording above says "upstream" rather than
+ * "released". #2686 documented seven rules the linter shipped on 2026-08-30,
+ * but npm `latest` is 0.5.1 (published 2026-08-27) and package.json pins
+ * `^0.5.1` - so five of those names were true statements about the linter and
+ * still unknown here, and check:skills went red on main for every pull
+ * request until they were listed. */
 const PENDING = new Map([
   // Empty, and it stayed empty for about an hour. `frozen-view-builder` was
   // listed here while this repository pinned 0.1.1; the bump to 0.2.0 made the
   // rule known, this gate reported the entry as stale on the same commit, and
   // it came out. That is the mechanism working, not an accident of timing.
+  //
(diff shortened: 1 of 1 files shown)


oblomov-dev:
Superseded by [#2689](https://github.com/abap2UI5/abap2UI5/pull/2689), and **this PR must not be merged now — it would turn `main` red.**

#2689 fixed the same failure the better way: instead of exempting the five names, it pins the linter to the commit the skills were written against, so the rules become genuinely *known* and their `Linter:` lines are actually verified rather than waved through. Measured on `2ad8c16`:

```
KNOWN  class-constructor-visibility      KNOWN  value-header-default-reassigned
KNOWN  into-corresponding-inline-decl    KNOWN  validating-setter-out-of-range
KNOWN  absent-boolean-overrides-default          total rules: 111  (was 93)
```

`npm run gates` on `main` is now **30 checked - all OK**.

That is exactly what makes this PR harmful rather than merely redundant. `skill-rule-gate.mjs` fails on a `PENDING` entry whose rule the pinned linter has — *"remove the entry — it exempts nothing and hides the next one"* — so all five entries here are now stale by that rule's own definition, and merging them would fail `check:skills` again. The two changes also conflict in that file.

Closing. Nothing to port: the problem it addressed is gone, and the mechanism it used is the one #2689 made unnecessary.

---

PR #2716, opened 2026-09-04
A message box shows its details, instead of hiding them behind a link

oblomov-dev:
# Description

`sap.m.MessageBox` renders `details` as a VBox of `[ message, "Show details" link, FormattedText(visible=false) ]` and offers **no option that opens it** — the link's press handler is the only thing that sets the text visible. That made sense while details were a stack trace nobody wants by default. It stopped making sense with #2712: what `message_box_display( )` renders out of a table, a structure, a tree or an object **is** the message, and a rendering behind a link the user has to find first is not shown at all.

So the frontend unfolds it. `showBox( )` gives a box that carries details an id (`MessageBox` has no `id` in the ABAP signature, so one is never sent from an app — and an id the caller did set is kept), and right after `show( )` walks the layout MessageBox built one statement earlier: the `FormattedText` is set visible, the link that would have revealed it is not. Every call is a public control API (`getContent` / `getItems` / `setVisible` / `isA`) and every step is guarded, so a release that lays the box out differently leaves the details collapsed — the behaviour before this — rather than throwing inside the display path.

**That put weight on content the sanitizer was throwing away.** `Lib.sanitizeMessageDetails` collected the top-level `<li>`s and took each one's `textContent`, so a whole subtree — a table inside a structure inside a row — arrived as one run-on line. Behind the link that was hard to see; as the content of the box it is the whole output. It now rebuilds the markup recursively and stays safe **by reconstruction rather than by removal**:

- every text node is escaped through a DOM property
- every element is re-emitted only if its tag is on the whitelist, and never with an attribute
- an element that is not on it contributes its children and nothing else
- `script` / `style` / `template` are dropped **with their content**, because dropping only their tag prints their source as text (a leading `<script>` is parked in `<head>` by the HTML parser and never reached the walk; one inside a `<li>` does)

The whitelist is the intersection of what `ui5_data_box_format( )` emits and what `sap.m.FormattedText` keeps: `ul, ol, li, strong, em, p, br`.

## `c_escape_html( )` from the catalogue, instead of a second name for it

The second commit is separate and small. The data renderer escaped through a private `html_escape( )` handling `&`, `<` and `>`; abap-util — the master catalogue `z2ui5_cl_ui5_util_context` is vendored from — already carries `c_escape_html( )`, which handles those three and the quote pair on top. AGENTS.md's "Utilities" says to look in the catalogue first and copy what is there, so the catalogue's method arrives under the catalogue's name and with the catalogue's body. The quote pair is not cosmetic: a value that lands inside an attribute must not be able to close it.

A body-level diff of all three vendored classes against abap-util came out of the same pass. What the catalogue is missing from here — the `ui5_data_box_format( )` closure, a CRLF/header-injection fix in `set_header_field( )`, a client leak in `client_call( )`, `rtti_check_ref_data( )` answering `abap_true` for an object reference, and about ten more — is a sync **into** abap-util and not part of this PR.

## How to test

`Z2UI5_CL_SMP_APP_502` in [abap2UI5/samples](https://github.com/abap2UI5/samples/pull/826) is the worked example: its *Structure*, *Table*, *Row limit* and *Depth limit* rows open boxes whose whole content is the rendering. Without this PR they open with a "Show details" link and, once clicked, show the nested rows as one run-on line.

Minimal check without the samples repository, in any app:

```abap
DATA(lt_msg) = VALUE bapiret2_t( ( type = `E` message = `first` )
                                 ( type = `W` message = `second` ) ).
client->message_box_display( lt_msg ).
```

Both texts stand in the box as a bullet list, with no link to press.

## Checklist

- [x] `npm run verify` passes (`npm run verify:full` when `app/webapp/` changed) — `verify:full` run; it stops at `npm run gates` on the **pre-existing** `check:shared` failure (`docs/agents/building-apps.md` differs from the app-template copy, and six `check-family-nav` consumers are gone upstream), which is red on `main` unchanged. Everything else passed: `check`, `check:standard`, `check:cloud`, `check:ui5` (0/0), `check:abap2ui5`, `check:format`, `check:eslint`, the app lint, and the build pipeline run on its own — downport → transpile → 1231 ABAP unit tests → 1071 JS specs → `check:app2abap` → `check:frontend`
- [x] Unit tests added/updated where it makes sense (`.testclasses.abap` / `node/tests/`) — four specs on the unfolding including the degrade-safely guard, the browser sanitizer specs now pin the nesting / the kept tags / a script inside a list item, and `ltcl_string->test_escape_html`
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` (see AGENTS.md) — `src/01/03` regenerated with `npm run app2abap`
- [x] Public API in `src/02/` only changed additively — `src/02/` is untouched

https://claude.ai/code/session_01KXeX2vwvHaEJwHeJ8akxra

---
_Generated by [Claude Code](https://claude.ai/code/session_01KXeX2vwvHaEJwHeJ8akxra)_


oblomov-dev:
**`check_gates` is red, and it is not this PR's.** The failing step is `shared-file-gate` with the same 7 problems it reports on `main`: run 220 on `a9f67ff` (`Fix SLIN "Redundant conversion for type STRING" in util_context`, #2713) failed the same way, before this branch existed.

```
  first difference at line 543
samples / samples-controls / samples-stack:
  .github/workflows/check-family-nav.yaml — declared a consumer, but the copy is gone upstream
  scripts/check-family-nav.mjs            — same
```

Worth stating explicitly, because the file is one this PR touches: the drift at line 543 is the `message_box_display( )` block #2712 added, not the sentence added here. My edit lands inside the same paragraph, three lines further down, and does not create the finding — the gate names the FIRST differing line and that line is unchanged on this branch. Nor can it fix it: the copy in `abap2UI5/app-template` has to be synced there, and the six `check-family-nav` entries need their consumers removed from `.github/shared/`'s manifest. Both are their own change, in another repository or in the manifest, and neither belongs in a PR about a message box.

Every other check on this head is green: `abaplint`, `test_unit` (1231 ABAP unit tests), `test_unit_js`, `test_node`, `browser` on chromium / firefox / webkit **and the pinned OpenUI5 1.71 leg**, `transpile`, `build`, `UI5_2X`, `render-gate`, CodeQL.

`check_app2abap` was cancelled during `npm --prefix app ci` — before any check ran — and has been re-run once; that is the drift gate over the regenerated `src/01/03`, and `npm run check:app2abap` reports no drift locally.

---

PR #2772, opened 2026-09-19
Four seams for hosting the framework outside SAP, and the transpiled runtime as a package

oblomov-dev:
# Description

abap2UI5 already runs on Node (`npm run express`). Hosting it in anything *else* — a CAP plugin is the case in hand — needed four small openings in the framework, none of them CAP-specific, plus a way to consume `node/output` without cloning and transpiling this repository. All five are additive; nothing that works today changes.

**Commit 3 is a bug fix and is the one to read first.**

### 1. `z2ui5_if_ui5_draft_store` — the draft persistence behind an interface

Swapped with `z2ui5_cl_ui5_srv_draft=>set_instance( )`. Session state is the one thing a host outside an SAP system cannot bring with it: the shipped store writes `Z2UI5_T_01` through Open SQL, which a NodeJS host satisfies with a private SQLite that no other part of that host can see or transact with. An implementation supplies the seven methods and decides where the rows live.

The owner-binding contract is in the interface's ABAP Doc, including the part an implementation must not get wrong: a read of somebody else's draft answers the same `NO_DRAFT_ENTRY_OF_PREVIOUS_REQUEST_FOUND` a missing one does, so a caller cannot tell a foreign draft from an absent one.

No `ALIASES` (`no_aliases` is an error here); every caller goes through `get_instance( )`, which hands back a fresh instance of the shipped store per call — byte-identical to what the direct calls did.

### 2. `z2ui5_if_ui5_serializer` — the app-state serialization behind an interface

`all_xml_stringify` / `all_xml_parse` on `z2ui5_cl_ui5_app_cont`, swapped with `set_serializer( )`. Same problem, other half: the default is `CALL TRANSFORMATION id` over S-RTTI, which has no counterpart outside ABAP. What an implementation owes is a **round trip, not a format** — `parse( stringify( container ) )` must answer a container the framework can go on with.

Both ends are `REF TO object` rather than `REF TO z2ui5_cl_ui5_app_cont`: an interface here may not reference a class (`intf_referencing_clas`) and naming it would close a cycle, since the container is what calls the interface. `narrow( )` casts once. Note the typed local in `parse( )` — the transformation rebuilds from the class named in the asXML and needs a concretely typed target.

### 3. The guarded codepage fallback — **a bug fix**

`z2ui5_cl_ui5_util_context=>conv_get_string_by_xstring` / `conv_get_xstring_by_string` reach `cl_abap_conv_codepage` first and the classic `CL_ABAP_CONV_*` pair second, but the fallback was raised **from inside the first CATCH**. On a runtime where neither exists, the second failure escaped as an unhandled exception from within an exception handler instead of as `UNSUPPORTED_CODEPAGE_API`.

The caller that pays for it is `z2ui5_cl_ui5_view_builder`, which builds its control-character escape set at class level: the failure took the view build down before any app code ran, and a host cannot catch what never became a class-based exception.

Now: the fallback is its own `TRY`, both failures are chained into `UNSUPPORTED_CODEPAGE_API`, and the builder degrades with `CLEAR gv_escape_controls` — it stops escaping control characters and renders the view.

**Found on a transpiled NodeJS runtime, where it presented as a HANG rather than an error.** A downstream project's transpiled core blocked on its first roundtrip for weeks while every dashboard stayed green, because the shape only exists where the codepage class is absent and no CI runs there. Reproduction and the red-to-green evidence: `test/core-runnable.test.js` in [cap2UI5/builder-abap2UI5-js#29](https://github.com/cap2UI5/builder-abap2UI5-js/pull/29).

### 4. `c_protocol` — the wire carries its own version

Every response stamps `S_FRONT.PROTOCOL` from `z2ui5_if_ui5_types=>c_protocol`, and `app/webapp/core/Server.js` compares it against its own `PROTOCOL` before reading anything else.

It is not the product version and does not move with a release; it moves when a response can no longer be read by a frontend written for the previous number — **the `S_ACTION` envelope replacing `S_FRONT.PARAMS` was such a change, and is the reason this exists.** The two halves ship together here and not everywhere: a port, a pinned webapp or a third-party shell can pair a backend and a frontend of different ages, and the failure mode was the worst available — the frontend looked for a key the backend no longer wrote and rendered its empty result.

A mismatch is reported to the user; a response *without* the field is let through, because a backend older than the field cannot be told apart from one that is merely older. AGENTS.md says to raise both halves in the change that breaks the wire.

### 5. `@abap2ui5/runtime` — the same build, also as an npm package

> **Reworked after `#2769` landed.** This started as a job of its own in `release.yaml`, written before `backend-prebuilt.yaml` existed. It does not need one: that workflow already downports, transpiles and proves the tree with the unit suite for the release tarball, so the package is now packed from **the same build**, as two steps at the end of that job. A second job would have spent another half hour producing the same bytes.

Two deliveries of one build, for different consumers:

| | `backend-<version>.tar.gz` (#2769) | `@abap2ui5/runtime` (here) |
|---|---|---|
| resolved by | name, from the GitHub release | `npm i`, from package.json |
| carries | `node/downport`, `node/output`, `node/deps` | `node/output`, `node/setup/setup.mjs`, **`app/webapp`** |
| for | a tool that downloads and builds against it (`mcp-server`) | a host that *runs* the framework — a CAP plugin, a serverless function |

The difference that matters is the **frontend**. A host that runs the framework serves the UI5 shell too, and pinning `@abap2ui5/runtime@X.Y.Z` is what gets it the backend, the frontend and `c_protocol` from one commit — the guarantee the wire version alone cannot give a host that pairs the halves itself. The tarball does not carry `app/webapp`, and an ordinary Node project would otherwise have to write and maintain its own fetch-and-unpack step for a release asset.

The manifest is **`node/setup/runtime.package.json`**, copied to `package.json` in a staging directory outside the checkout at pack time. It is deliberately *not* `node/package.json`: a `package.json` inside `node/` makes that directory an npm package root, so `npm run <script>` from there stops finding this repository's scripts — `cd node && npm run express` answers *"Missing script"*, which is how `node/playwright.config.js` starts its web server, and all four browser projects fail to boot. CI found that; the file's own header now records it. The committed `0.0.0-set-at-release` is deliberate, and there is no `type` field on purpose — the playwright configs are CommonJS.

The `.tgz` is uploaded as a workflow artefact on every run; `npm publish` runs only when the organisation has an `NPM_TOKEN`, warning loudly and staying green without one — and the release was created *before* this workflow started, so neither a registry failure nor a missing token can touch it. Dry run: 1,307 files, 1.5 MB packed, 16.4 MB unpacked.

**This one needs a decision, not just a review.** If the organisation does not want to publish to npm, drop that part and the four seams stand on their own.

## How to test

```bash
npm run verify              # the whole thing
npm run unit                # the transpiled ABAP unit suite alone
```

Seam 1 and 2 have no behaviour to test on a system — the defaults are the previous code paths, and `get_instance( )` / the default serializer are what every existing test already exercises. Seam 3 is only reachable where `cl_abap_conv_codepage` and `CL_ABAP_CONV_*` are both absent, which is why its evidence is a downstream runtime rather than a test here; on a system nothing about it changes. Seam 4 is visible in any response:

```bash
npm run express
curl -s -X POST localhost:3000/sap/bc/z2ui5 -H 'Content-Type: application/json' \
  -d '{"value":{"S_FRONT":{"APP":"z2ui5_cl_ui5_app_hi_world","SEARCH":"?app_start=z2ui5_cl_ui5_app_hi_world"},"MODEL":{}}}' | head -c 120
# → {"S_FRONT":{"APP":"...","ID":"...","PROTOCOL":2, ...
```

A host swapping seam 1 and 2, with the drafts in a CDS entity and the state surviving a SIGKILL, is in [cap2UI5/cap2UI5#72](https://github.com/cap2UI5/cap2UI5/pull/72).

## Checklist

- [x] `npm run verify` and CI both pass, **with two pre-existing exceptions**, each verified to fail identically on `main` in the same checkout:
  - `check_gates` → `shared-file-gate` — 4 differences against `app-template` and the three samples repositories (a `sync-shared.yaml` block and one line in `building-apps.md`). Names only repositories this branch does not touch.
  - `test_unit_js` → `node/tests/devtoolsConsole.spec.js › bounds the nodes of a map-shaped object` — 1 of 1205. Reproduced on a bare `origin/main` checkout here (28 passed, 1 failed in that file alone), so it is either a `main` breakage or environment-specific; this branch touches nothing in the console capture.

  Green on CI at `0cae842`: all four `browser` projects and `test_browser`, `test_node`, `test_unit`, `transpile`, `check_app2abap`, `render-gate`, `analyze`, `build`, `UI5_2X`, CodeQL and both abaplint reports.
- [x] Unit tests added/updated where it makes sense — see the note under *How to test*: the three ABAP seams have no new behaviour to assert (the defaults are the previous code paths), and seam 3's failure mode is only reachable on a runtime this repository's CI does not have. The regression test for it lives where that runtime is.
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` — `src/01/03` is regenerated by `npm run app2abap`, as the `c_protocol` change to `app/webapp/core/Server.js` requires (and again after the merge, which brought upstream's own frontend changes).
- [x] Public API in `src/02/` only changed additively — `z2ui5_cl_ui5_view_builder` gained a guard around an existing class-level build; no signature moved.
- [x] `changelog.txt` has lines under `unreleased` — five, one per change.
- [x] Changes were made via abapGit: **no** — written as files and checked with `npm run check:abapgit` (BOM on the new `.xml` sidecars, LF, single EOF newline, `<CLSNAME>` matching the filename) and `npm run rename`, which is what caught the two object names that were over the 25-character namespace budget.

## Note on provenance

This branch was written by Claude Code in a single session, including the analysis that found the codepage bug. Everything asserted above is measured in that session; where something is *not* proven — seam 3 on a real SAP system, for instance — it says so rather than implying a test that does not exist. Two defects of my own reached CI and are fixed in `88c78d5` and `0cae842`; the comments below say what they were and why nothing local saw them.

https://claude.ai/code/session_01JoANwtaK8Jm9srGEqSxmUF


oblomov-dev:
Rebased onto current `main` (merge commit `de2679d`) — the branch was conflicted against the three commits that landed while this was being written. And one of them changed the answer to part 5, so I reworked it rather than leaving a duplicate in.

**`#2769` landed ~15 minutes before I opened this**, and it already attaches the transpiled backend to every release. My `runtime` job in `release.yaml` would have run its own `downport` + `auto_transpile` to produce the same bytes `backend-prebuilt.yaml` had just produced — half an hour of runner time for nothing. The npm pack and publish are now **two steps at the end of that job**, reusing its build, and `release.yaml` is back to what it was.

What is left after the rework is a genuinely different delivery rather than a second copy of the same one:

- the tarball carries `node/downport` / `node/output` / `node/deps` and is resolved by name from a release — what `mcp-server` builds against;
- the package carries `node/output`, the setup hook and **`app/webapp`**, and is resolved by `npm i` — what a host that *runs* the framework needs, because it serves the UI5 shell too.

The frontend is the part that cannot come from the tarball, and it is exactly the part `c_protocol` (seam 4) exists to protect. Pinning one version gets a host both halves from one commit.

Conflicts resolved: `.gitignore` (both blocks kept), `changelog.txt` (both sets of entries, upstream's first), and `src/01/03/z2ui5_cl_ui5f_preload.clas.abap`, which is generated — I took upstream's side and re-ran `npm run app2abap` rather than hand-merging it.

Re-verified on the merged tree: 31 of 32 gates, abaplint 0 issues, downport, transpile, the ABAP unit suite, `check:js`, `check:app2abap`, `check:frontend`. Two checks fail, both reproduced on a bare `origin/main` checkout in the same environment and named in the checklist above — `check:shared` (4 drifts against repositories this branch does not touch) and one frontend test in the console capture, which this branch does not touch either.

The workflows on this PR have not started; only the abaplint app has reported. If they are waiting on a maintainer's approval for a first-time contributor, a nudge would be welcome — I would rather have CI's verdict than my own.

---

oblomov-dev:
CI ran, and it caught something none of the checks I ran locally could. Fixed in `88c78d5`.

## `abaplint` — mine, and a good catch

`npm run rename` stopped with a bare `Error: Name not allowed`, no file named. There is a **fourth target** besides 7.02, standard and cloud — the `rename_<name>` branches — and it substitutes `z2ui5` with a namespace of up to ten characters. My two new objects were 27 characters:

```
z2ui5_if_ui5_app_serializer (27) → znamespace_if_ui5_app_serializer (32)   ✗ ABAP allows 30
```

The budget is **25 characters**, and it is written down — in the header of `.github/abaplint/rename.jsonc`, the file that enforces it. I should have read it before adding an object; the two serializer names were the only ones in the tree over it.

They are now `z2ui5_if_ui5_serializer` / `z2ui5_cl_ui5_serializer` (23), which reads better anyway: what is serialized is named by the parameter, not by a word in the type name.

Worth recording why nothing local saw it: `check:naming` reads the namespace segment and not the length, `check:abapgit` checks that `<CLSNAME>` matches the file name and not how long either is, and abaplint over `src/` is green because the name is perfectly legal *until* it is renamed. Only `npm run rename` decides it, which is why `abaplint.yaml` runs it last. The `abap-check` skill gains the case in section 4, with that list — its own rule for a defect that reached a branch.

Re-verified: `rename` applies cleanly (0 issues, 363 files), abaplint 0 issues, `check:abapgit` OK (BOM, LF, single EOF newline and `<CLSNAME>` on both renamed sidecars), plus `check:naming`, `check:atc`, `check_visibility`, `check:dynamic`, `check:api` and `skill-rule`.

## The other two red checks are not this branch's

Both were already named in the PR checklist; CI now confirms them from its own side.

- **`check_gates`** fails on `shared-file-gate`: 4 drifts against `app-template` and the three samples repositories (a `sync-shared.yaml` block and one line in `building-apps.md`). It fails identically on `main` — verified by checking `main` out and running it — and names only repositories this branch does not touch. The `linter-mirror` step in the same job printed *"the linter's MAIN already matches this tree … Not a failure"*, so that one is fine.
- **`test_unit_js`** fails on one of 1205 frontend tests, `devtoolsConsole.spec.js › bounds the nodes of a map-shaped object`. Reproduced on a bare `origin/main` checkout here (28 passed, 1 failed in that file alone), so it is either a `main` breakage or environment-specific; either way this branch touches nothing in the console capture.

I have not re-run either — they are deterministic and I have the reproduction on `main`, which is stronger evidence than a second identical run. If you would rather I did, say so.

---

oblomov-dev:
Second one CI found that I could not, and it is the sharper of the two. Fixed in `0cae842`.

## All four browser projects failed to boot

```
[WebServer] npm error Missing script: "express"
Error: Process from config.webServer was not able to start. Exit code: 1
```

`node/playwright.config.js` starts its web server with `npm run express` **from `node/`**. A `package.json` inside that directory makes it an npm package root, so `npm run` stops walking up to this repository's scripts and finds none. The manifest I added for `@abap2ui5/runtime` was that `package.json`.

Why nothing here saw it: the manifest is not ABAP, not a workflow and not a source file, so no gate reads it — and `npm run express` from the **root** is unaffected. Only a caller inside `node/` breaks, and the browser suite is the one caller there is. I had run `npm run verify` and the build pipeline; neither starts a browser.

**The fix:** the manifest is `node/setup/runtime.package.json` now — a plain data file whose header records why it may not be called `package.json` — and `backend-prebuilt.yaml` assembles the package in a **staging directory outside the checkout**: the manifest as `package.json`, with `node/output`, `node/setup/setup.mjs` and `app/webapp` beside it, and `npm pack` there. Nothing is copied into the checkout any more, so `/node/webapp/` leaves `.gitignore`.

Verified: `cd node && npm run express` answers 200 on `/ping` again, and the staging pack produces the same artefact as before — 1,307 files, 1.5 MB packed, 16.4 MB unpacked, version read from `package.json`, with `output/`, `setup/setup.mjs` and `webapp/` in the relative positions `output/init.mjs` imports.

**What is *not* proven locally:** the browser suite itself. It needs `https://sdk.openui5.org`, which this sandbox has no route to — so what I reproduced and fixed is the boot that failed, not the tests behind it. CI on this head is the first run that can say.

## Where the two PRs stand

| head `0cae842` | |
|---|---|
| `abaplint` (namespace rename) | fixed in `88c78d5` — object names back inside the 25-character budget |
| `browser (×4)`, `test_browser` | fixed here |
| `check_gates` → `shared-file-gate` | **not this branch's** — 4 drifts against `app-template` and the samples repositories, fails identically on `main` |
| `test_unit_js` → `devtoolsConsole › bounds the nodes of a map-shaped object` | **not this branch's** — 1 of 1205, reproduced on a bare `origin/main` checkout here |

Two mistakes of mine in one afternoon, both of the same shape: something true about this repository that is written down where I did not look — the 25-character budget in `rename.jsonc`'s header, and `playwright.config.js` starting its server from inside `node/`. Each is now recorded next to the thing that enforces it, which is the only version of "won't happen again" I can actually ship.

---

oblomov-dev:
CI on `0cae842` confirms both fixes — including the browser suite, which is the part I could not run here.

**20 of 22 checks green**, among them all four `browser` projects and `test_browser` (*"every browser leg passed"*), `test_node` — which boots with the `npm run express` that the manifest had broken — plus `test_unit`, `transpile`, `check_app2abap`, `render-gate`, `analyze`, `build`, `UI5_2X`, CodeQL and both abaplint reports. `abaplint / cross check` is still running.

The two that remain red are the ones established as not this branch's, unchanged: `check_gates` → `shared-file-gate` (4 drifts against `app-template` and the samples repositories) and `test_unit_js` → `devtoolsConsole › bounds the nodes of a map-shaped object`. Both reproduce on `main` in the same checkout.

Ready for review from my side.

---

oblomov-dev:
I said twice that the two red checks "fail identically on `main`" and left it there. That was true but not useful — it says whose they are without saying what they are. Both are now **root-caused**, and one of them has a patch below.

Neither is this branch's: it touches none of the four files involved (`git diff --name-only origin/main...HEAD` — no `building-apps.md`, no `sync-shared.yaml`, no `devtools/Console.js`, no `devtoolsConsole.spec.js`).

## 1. `test_unit_js` → `devtoolsConsole › bounds the nodes of a map-shaped object`

**The test was merged red, by `1f2cccb` (#2771), which is `main`'s head.** `git log -- node/tests/devtoolsConsole.spec.js` and `-- app/webapp/devtools/Console.js` both end at that commit, and it added `MAX_NODES` *and* this test together. It has therefore never passed — this is not a flake and not environment-specific, which is what I could not rule out before.

The cause is that the third assertion asserts something a `JSON.stringify` replacer cannot do:

```js
expect(text).not.toContain(`"k${max * 3 - 1}"`);
```

A replacer answers a key's **value**. It cannot remove the key. So for a map-shaped object every one of the 3,000 keys is emitted whatever the node budget does — the bounded ones simply carry `"[...]"` instead of their object:

```
…,"k2997":"[...]","k2998":"[...]","k2999":"[...]"}
```

`"k2999"` is present, the assertion fails. The sibling test one block up ("bounds a long array") passes because an **array** is bounded by the `MAX_ITEMS` slice, which really does drop the items.

The implementation is doing what #2771's own commit message claims for it — *"MAX_NODES caps the walk"*. Measured on that test's own input (3,000 × `{i}`) and on the shape the code comment names (a lookup of rows):

| | without the cap | with it |
|---|---|---|
| 3,000 × `{i}` (the test's input) | 54,781 chars | 48,779 — **11%** |
| 3,000 × `{id, name, city, amount}` | 205,930 chars | 99,878 — **52%** |

999 objects are expanded instead of 3,000, which is the walk being capped. The test just asks for a different property than the one the cap has.

### Proposed patch

Assert the property the cap actually has — the value is the marker, not the object:

```diff
     expect(text).toContain("[...]");
-    expect(text).not.toContain(`"k${max * 3 - 1}"`);
+    // The last key is still EMITTED - a replacer answers a key's value and
+    // cannot remove the key - but its value is the marker, not the object.
+    expect(text).toContain(`"k${max * 3 - 1}":"[...]"`);
+    expect(text).not.toContain(`"k${max * 3 - 1}":{`);
```

Verified here: **29/29 green** in that file with the patch, and it **discriminates** — disabling the `++nodes > MAX_NODES` branch turns it red again (`1 failed, 28 passed`), so it still pins the cap rather than merely passing.

I have deliberately **not** put this on this branch. It is a frontend test in code these four seams do not touch, and carrying an unrelated fix in here would bury it in a 27-file diff. Say the word and I will push it — or take the diff, it is three lines.

One thing worth deciding separately, not proposed as part of this: the cap bounds the **walk**, not the **output**. A 3,000-key lookup still produces ~49 KB before `MAX_TEXT_CHARS` cuts it, because 3,000 `"[...]"` markers are about as long as the values they replace. If the intent of #2771 was to bound the string as well, an object needs the same treatment arrays get — a head of `MAX_ITEMS` keys plus one `"… N more"` marker — which would *also* make the original assertion true as written. That is a behaviour change, so it is yours to want, not mine to assume.

## 2. `check_gates` → `shared-file-gate`

Not a mystery either — **#2771 predicted one half of it in its own commit message**:

> building-apps.md described check_queue_last and never named check_no_busy … The app-template mirror of the guide regenerates from here (`npm run agents` there), so **check:shared stays red until that follow-up lands** — the same sequence as #2758.

`npm run check:shared` here reports exactly 4 problems, and they split in two:

| drift | source | since |
|---|---|---|
| `docs/agents/building-apps.md` → `app-template/AGENTS.md`, line 565 (the `check_no_busy` paragraph) | #2771 | `main`'s head, follow-up expected |
| `.github/shared/sync-shared.yaml` → `samples`, `samples-controls`, `samples-stack`, line 32 (a `permissions:` block here, absent there) | **#2719, 2026-09-05** | two weeks before this branch existed |

The second is the larger share and is older than this PR by a fortnight. Those three repositories re-sync on a weekly cron (`47 5 * * 2`), so the gate goes green when their next sync run picks up the `permissions:` block — nothing to fix in *this* repository, where the file is already the source of truth. I cannot verify that from here: `samples`, `samples-controls`, `samples-stack` and `app-template` are outside this session's repository access.

## Where that leaves this PR

Unchanged on the merits — 20 of 22 checks green at `0cae842`, the two red ones established as `main`'s and now explained rather than merely attributed. Still ready for review, and still needing the decision in part 5 (publish `@abap2ui5/runtime` to npm, or drop that part and keep the four seams).

---

oblomov-dev:
**Correction — half of `check_gates` was mine all along.** Fixed in `72b8da9`.

`check_gates` has **two** failing steps. I looked at `shared-file-gate`, established it fails identically on `main`, wrote *"not this branch's"* and stopped — at `0cae842`, and again in this morning's root-cause comment. The other step is mine, and has been red since `88c78d5`:

```
Prose naming an object this repository does not ship:

  .claude/skills/abap-check/SKILL.md:808  z2ui5_if_ui5_app_serializer
  .claude/skills/abap-check/SKILL.md:808  z2ui5_cl_ui5_app_serializer
```

It is the abap-check entry I wrote *about* the 25-character rename budget — and it names the old spelling, because there the defect **is** the name: at 27 characters those two were the only objects in the tree over the budget. `prose-name-gate` is right to stop on them; a reader who looks one up finds nothing and cannot tell a rename from a removal.

The gate has the mechanism for precisely this and I did not use it. `HISTORICAL_IN` is a **per-file** allow-list whose own comment reads:

> the abap-check catalogue is a list of defects that shipped, so it names the objects as they were named when they broke

Both names are registered there now, for that file only, with the reason. Verified: `prose-name: 51 file(s) checked, every name resolves - OK (10 external names allowed)` — and it still **discriminates**, a bogus `z2ui5_cl_ui5_does_not_exist` dropped into `AGENTS.md` is still reported, so the registration excuses those two names in that one file and nothing else.

### Where `check_gates` stands now

| step | whose |
|---|---|
| `prose-name-gate` | **mine** — fixed in `72b8da9` |
| `shared-file-gate` | `main`'s — 4 drifts, unchanged; one predicted by #2771's own commit message, three from #2719 (2026-09-05) against the samples repositories, which re-sync on their weekly cron |

`test_unit_js` is unchanged and unaffected: `devtoolsConsole › bounds the nodes of a map-shaped object`, merged red by `1f2cccb` (`main`'s head), patch proposed in [my earlier comment](https://github.com/abap2UI5/abap2UI5/pull/2772#issuecomment-5748688700).

The lesson, since this PR's history is its evidence: **establishing that one failing step belongs to somebody else is not the same as reading the job.** A job is red for as many reasons as it has steps, and the one I did not go looking for is the one that was mine.

---

PR #2773, opened 2026-09-20
devtoolsConsole: assert what MAX_NODES bounds, not what a replacer cannot do

oblomov-dev:
# Description

`node/tests/devtoolsConsole.spec.js › bounds the nodes of a map-shaped object` **has never passed.** #2771 added `MAX_NODES` and this test in the same commit, so `test_unit_js` has been 1204/1205 since that merge — not a flake, not environment-specific, and still red on `main` at `e75cfd7`.

The third assertion asks for something a `JSON.stringify` replacer cannot do:

```js
expect(text).not.toContain(`"k${max * 3 - 1}"`);
```

A replacer answers a key's **value**. It cannot remove the key. So for a map-shaped object every one of the 3,000 keys is emitted whatever the node budget does — the bounded ones simply carry `"[...]"` where their object was:

```
…,"k2997":"[...]","k2998":"[...]","k2999":"[...]"}
```

`"k2999"` is present, the assertion fails. The sibling test one block up passes because an **array** is bounded by the `MAX_ITEMS` slice, which really does drop items — which is why its assertion has a different shape.

**The implementation is doing what #2771 claims for it**, *"MAX_NODES caps the walk"*, and the cap is worth having: 999 objects expanded instead of 3,000.

| | without the cap | with it |
|---|---|---|
| 3,000 × `{i}` (the test's own input) | 54,781 chars | 48,779 — **11%** |
| 3,000 × `{id, name, city, amount}` (the shape the code comment names) | 205,930 chars | 99,878 — **52%** |

So the fix is the **assertion**, not the code: the last key's *value* is the marker rather than the object.

## How to test

```bash
npm run check:js     # 1205 passed — the first fully green frontend run
```

Verified to **discriminate**: disabling the `++nodes > MAX_NODES` branch in `app/webapp/devtools/Console.js` turns it red again (1 failed, 28 passed), so it still pins the cap rather than merely passing.

## Left alone deliberately — a question rather than a patch

The cap bounds the **walk**, not the **output**. A 3,000-key lookup still produces ~49 KB before `MAX_TEXT_CHARS` cuts it, because 3,000 `"[...]"` markers are about as long as the values they replace — that is the 11% row above. Giving objects the head-plus-marker treatment arrays already get would bound the string too, and would make the original assertion true as written. That is a behaviour change, so it is not in here.

## Checklist

- [x] `npm run verify` passes — `check:js` 1205/1205; `prose-name`, `spec-inventory`, `assertion` and `check:eslint` green on this diff
- [x] Unit tests added/updated where it makes sense — this **is** the test change; no production code is touched
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` — nothing under `src/` at all
- [x] Public API in `src/02/` only changed additively — unchanged
- [x] `changelog.txt` has a line under `unreleased` when this changes behaviour — **not applicable and gate-confirmed**: `changelog-entry-gate` answers *"the diff does not touch src/02/** or the api snapshot - no entry required"*. No behaviour changes; a test starts measuring the property the code has.
- [x] Changes were made via abapGit: **no** — one JavaScript test file, edited as a file

https://claude.ai/code/session_01JoANwtaK8Jm9srGEqSxmUF

---
_Generated by [Claude Code](https://claude.ai/code/session_01JoANwtaK8Jm9srGEqSxmUF)_


oblomov-dev:
**`check_gates` is red, and it is not this PR's.** Naming it here rather than leaving it to be re-diagnosed.

This diff is one JavaScript test file (+8 −1). The whole job has exactly **one** `##[error]`, and it is `shared-file-gate` with the same four drifts `main` carries:

| drift | introduced by | |
|---|---|---|
| `docs/agents/building-apps.md` → `app-template/AGENTS.md`, line 565 | **#2771** | its own commit message predicts it: *"the app-template mirror of the guide regenerates from here … so check:shared stays red until that follow-up lands"* |
| `.github/shared/sync-shared.yaml` → `samples`, `samples-controls`, `samples-stack`, line 32 (a `permissions:` block here, absent there) | **#2719, 2026-09-05** | those repositories re-sync on a weekly cron (`47 5 * * 2`) |

The second group is the larger share and predates this branch by two weeks. Neither is fixable from this repository: the file here is already the source of truth, and `app-template`, `samples`, `samples-controls` and `samples-stack` are outside this session's repository access, so I cannot open the follow-up either.

Every other step of `check_gates` is green on this head, including `prose-name-gate` — which was red on the previous branch and whose fix is now in `main`.

No re-run spent: the gate reads those repositories live from GitHub and they have not re-synced, so a second run produces the same four lines.

**`test_unit_js` is what this PR is for**, and it is the one check the diff moves: 1204/1205 → **1205/1205**.

---

PR #2775, opened 2026-09-22
CONTROL_BY_ID: register setCurrentStep so the Wizard gets a control, not a string

oblomov-dev:
# Description

Reported from a real system running samples-controls app 101 (`sap.m.sample.Wizard`): every failing validation keystroke logged

```
The given step could not be set as current step.
```

and the `setCurrentStep` frontend action did nothing.

Not a port defect — the port reproduces the original's `setCurrentStep( ProductInfoStep )` faithfully.

`setCurrentStep` was the one Wizard method missing from `CONTROL_METHODS`, while all three of its siblings are on it as `["controlId"]`:

```js
discardProgress: ["controlId"],
setNextStep:     ["controlId"],
goToStep:        ["controlId", "bool"],
//  setCurrentStep — absent
```

Without an entry the call takes the unlisted-method path: `kinds` is `null`, `castArgs` falls through to `castArgAuto`, and the raw backend string goes over unchanged. `sap.m.Wizard` then does

```js
var oStep = (typeof vStepId === "string") ? Element.getElementById(vStepId) : vStepId;
```

against ids that carry the runtime view prefix the backend never sees, finds nothing, and takes its `else` branch — `Log.error(...)` and no action.

A **silent no-op**: no wrong step, no exception, nothing for an app to catch. That is why it survived — the view renders, the wire is in the XML, the handler runs. Same shape as the `backToPage` gap that gave `pageId` its argument kind, on the same port one method over.

The fix is `setCurrentStep: ["controlId"]`, which resolves the control and hands the object over — what `setCurrentStep`'s own string-or-control ternary wanted all along.

`src/01/03/` regenerated with `npm run app2abap` (the whole pipeline, so the generated tree carries only `ctrlcall_js` and the `preload` `build_hash`, no normalization drift).

## How to test

The new `node/tests` case is a **regression** test, not merely a passing one — verified in both directions:

```
npx playwright test -c node/playwright-unit.config.js --grep "setCurrentStep"
```

With the registry line removed it fails with the raw id where the control belongs, which is the production bug exactly:

```
- Expected        + Received
  Array [ Array [ "current",
-     Object { "id": "STEP2" },
+     "STEP2",
  ] ]
```

With the line in place it passes, alongside the three sibling Wizard tests.

In an app: any `follow_up_action( val = cs_event-control_by_id t_arg = VALUE #( ( 'wizard' ) ( 'setCurrentStep' ) ( 'step2' ) ) )` — before, the browser console carried the UI5 error on every call and the wizard did not move; after, it moves.

## Checklist

- [x] `npm run verify` passes (`npm run verify:full` when `app/webapp/` changed) — **with one pre-existing failure, see below**
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` (see AGENTS.md)
- [x] Public API in `src/02/` only changed additively — not touched at all

`npm run verify:full` reports **1 of 32 gates failed: `check:shared`**. It is not this diff — it fails identically on a clean checkout (confirmed by stashing), and none of its four reported files are in the diff: `docs/agents/building-apps.md` vs the `app-template` copy, and `.github/shared/sync-shared.yaml` vs the `samples` / `samples-controls` copies. Downstream copies that have drifted; left for whoever owns that sync rather than folded in here.

Everything else in that run is green: `abaplint` 0 issues (267 files), `check:standard` and `check:cloud` 0 issues (363 files each), `check:format` 0 issues (265 files), eslint, `check:abap2ui5`, `check:ui5`, app lint. Full frontend suite separately: **1203 passed**.

https://claude.ai/code/session_01Dx6GCN3neQcrDarjPHkfF4

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dx6GCN3neQcrDarjPHkfF4)_


oblomov-dev:
`check_gates` is red on `da3ae31` with **two** failing steps. One is mine, one is not, and I want to be explicit about which is which — the last time this job was reported on, one failing step was established as somebody else's and the other went unread (see the final commit of #2772).

**`linter-mirror-gate` — mine.** Fixed in abap2UI5/linter#110, which has to merge first.

The gate distinguishes two cases and says so in its own output. On `main` (`09d07c9`) it reports three drifts — `z2ui5_t_02`, `HASH_BACK`, and the stale `CLIPBOARD_APP_STATE` / `WIZARD_SET_NEXT_STEP` — and **passes**:

> linter-mirror: the INSTALLED @abap2ui5/linter drifted, but the linter's MAIN already matches this tree — fixed upstream, PENDING the monthly linter release and the dep bump here. Not a failure: releases never gate a merge.

On this branch it reports the same three **plus** `setCurrentStep`, and **fails**:

> linter-mirror: @abap2ui5/linter mirrors something this change moved (exit 1). The linter's MAIN drifts too, so no fix is on its way on its own.

`CONTROL_BY_ID id-argument methods` and `CONTROL_BY_ID method names` both read `in sync` on `main` and both drift here — the two lists the one new whitelist entry feeds. So this is not the pre-existing noise it resembles; the gate is doing exactly what it was built for, and linter#110 is the fix it asks for.

**`shared-file-gate` — not this branch's.** Red on `main` with the identical four problems (`docs/agents/building-apps.md` vs the `app-template` copy, `.github/shared/sync-shared.yaml` vs the `samples` / `samples-controls` / `samples-stack` copies), and none of those files is in this diff. Verified locally too: `npm run check:shared` fails the same way on a clean checkout with nothing of mine applied. Not fixable from here — the copies live downstream.

No re-run requested: neither failure is a flake, and one of them is genuinely waiting on another merge.

**Also resolved on this push:** the PR was `mergeable_state: dirty` from the moment it was opened, because `main` moved to `09d07c9` (#2774) in between. That is why only the three abaplint.app checks had run and none of the Actions workflows — GitHub could not build the merge ref. `main` is merged in now (no rebase, no force-push), the conflict was the generated `build_hash` in `preload` and was resolved by regenerating rather than by hand, and all eleven checks are running.

---

PR #2780, opened 2026-09-23
Frontend review follow-up: per-component context, small hardenings, cc/core cleanup, devtools split

oblomov-dev:
# Description

Follow-up to the frontend review of 2026-09-23 (after #2777, #2778, #2779), in four commits. No public API change (`src/02` untouched), no new dependency, no protocol name changed.

**1. Instance guard and three small hardenings** (`563e058`)
- `Component._claimSingleInstance()` refused a second live `z2ui5.Component` on the same page - the stopgap that commit 4 removes again, once the state is per component.
- `OPEN_NEW_TAB` opens with `"noopener,noreferrer"` instead of clearing `opener` on the returned window.
- `cc/History` refuses a `search` value without a leading `?` (it rewrote the path).
- The state records keyed off the wire (`timers`, `shortcuts` and its scope maps, `viewSizeLimits`, `treeStates`) are prototype-less.

**2. Cleanup in core/ and cc/** (`33a0fb1`)
- `cc/Dirty` installs the unload prompt as a `beforeunload` listener; the assignment to `window.onbeforeunload` overwrote and then removed a host page's handler.
- `MultiInputExt`, `SmartMultiInputExt`, `UploadSetExt` keep their bound handlers and detach them from the target control in `exit()`.
- `cc/Info` reads the device model through its own `getModel` (every slot view carries it), so it works inside a popup.
- `cc/Geolocation` fires its `error` event with code 0 when the browser has no Geolocation API instead of staying silent.
- `model/formatter`: `DateAbapDateTimeToDateObject` no longer throws on a `null` time.
- `core/Session.reset()` exists and `Component.exit` calls it (the one module with send latches and no reset).
- Two seams: `ViewSlots.markChanged` (the one writer of the tracked model's change set outside `actions/Slots`; `cc/Scrolling` used the set by name) and `Env.controlFilters` (the `getFilters` since 1.96 / private `aFilters` fallback, moved out of `cc/UITableExt` per rule 8).

**3. Developer tools** (`027ecc4`)
- `Inspect.js` 1167 → 658 lines: log part to `Log.js`, bindings part to `Bindings.js`, help text to `Help.js`; `Recorder.js` 1070 → 907: the two diff algorithms to `Diff.js`. Inspect re-exports `formatLog`/`formatBindings`/`formatHelp`, so callers are unchanged.
- Shared once: `Persist.js` (the one sessionStorage access), `SlotXml.js` (the one slot XML reader, below Tabs/Inspect/Picker), `Format.section/renderValue/describeValue`.
- The devtools fragment id is derived from the control instance (`<id>--tools`) instead of the page-global `z2ui5DeveloperTools`; `AbapSource` builds its preview iframe as a DOM element and serializes it; `ErrorView`'s friendly-dialog fallback logs why the UI5 dialog failed.

**4. One frontend context per component** (`97d3eec`) - stage 2 of `backlog/items/embed-as-reuse-component.md`
- The frontend state was a module singleton (`core/AppState.js` plus the module-scoped records of Server, Session, Router, Shortcuts, ScrollFocus, ErrorView and the devtools), and `Component.init` reset it for whoever came last: a launchpad in keep-alive mode or a host creating the component twice corrupted the first instance, which failed later with "App Terminated".
- New `core/Context.js`: `Context.create(component)` builds one context per `z2ui5.Component` - the state (`AppState.createState()` is now only the shape) and a record per module (`ctx.server`, `ctx.session`, `ctx.router`, `ctx.shortcuts`, `ctx.scroll`, `ctx.errorView`, `ctx.devtools`). A module below the component takes the context as its first argument; an action handler reads it off the controller (`oController.ctx`, set by `App.controller`); a custom control asks `Context.of(this)`, resolved through the owner component (views and fragments are built under `Context.runAsOwner` in `actions/Slots`, so UI5's own `getOwnerComponentFor` answers for every control inside them) or the parent chain up to a slot view. `Component.exit` destroys its own context only. The instance guard is gone.
- What stays page-wide by nature, and deliberately: the URL hash, title and favicon, the global BusyIndicator, the messaging facade, the devtools console capture (use-counted), the `cc/Dirty` prompt and the raw fatal-error overlay. Listed in the header of `Context.js`; `Lib.logError`'s ring is a page-wide diagnostic for the same reason and is `Lib.errors` now.
- Frontend-internal API change for custom-control authors reaching into `core/` (changelog line under `unreleased`): the state is `Context.of(this)?.state` instead of `AppState.state`, and the `core/` functions take the context first. `Lib.hookCallback`, `Lib.afterRoundtrip` and the `ViewSlots.byIdOfOwner`/`trackedModel`/`markChanged` seams are unchanged.
- Proven in the browser by the new `node/tests/e2e/two-components.spec.js`: the served page's component plus a second one created through a `ComponentContainer`, both rendering their own main view under their own context, and the container teardown of the second leaving the first running. It also runs offline: `e2e/fixtures.js` with `UI5_PINNED_RESOURCES` (a local OpenUI5 `resources/` tree) additionally allows `'unsafe-eval'`, which an unbundled source tree needs.
- The unit specs load the modules against a spec context (`loadLibModule.js`: `specContext`, `bindContext`, `withSpecController`); `context.spec.js` is new, `appState.spec.js` covers only the shape now. AGENTS.md rule 8, `docs/agents/repository-map.md`, `docs/agents/test-inventory.md` and the backlog item (stage 2 done, state open) say what is per component and what is not.

`src/01/03/` regenerated with `npm run app2abap` in every commit (new carrier `z2ui5_cl_ui5f_context_js`); changelog lines under `unreleased`.

Left out on purpose, each a change of its own: a hash-source CSP instead of `'unsafe-inline'` for the GET page's inline script (ABAP, SHA-256 on all three targets, ETag cache), an exit switch for the developer tools (a new field on `z2ui5_if_ui5_exit`), and stage 1 of the embedding item (an embedded flag that turns the page-wide behaviours off) plus a wrapping custom control.

## How to test

- `npm run check:js` - 1290 specs, among them `context.spec.js` (create/of/destroy/runAsOwner), the ported core, cc and devtools specs, and the earlier ones for the hardenings and the cleanup.
- `cd node && npx playwright test --config=playwright.config.js --project=chromium tests/e2e/two-components.spec.js` against `npm run express` - both instances render, the second one's container teardown leaves the first alive.
- In a browser: create the component twice on one page (a `ComponentContainer` with `name: "z2ui5"`), each renders and answers its own roundtrips; FLP re-launch and page reload are unaffected.
- Devtools (Ctrl+F12): every tab renders as before; the Log, Bindings and Help tabs come from the new modules; Source preview still shows the class.
- A form with `z2ui5.cc.Dirty`: the "leave page?" prompt appears while dirty and disappears when clean; no `window.onbeforeunload` is assigned.

## Checklist

- [x] `npm run verify` passes (`npm run verify:full` when `app/webapp/` changed) - on the final tree: Prettier, ESLint (app and repository-wide `eslint .`), `tsc`, ui5lint 0/0, abaplint 0 (standard, cloud, format), `check:app2abap`, `check:modules`, `check:specs`, `check:backlog`, `check:prose`, `check:conventions`, `check:frontend`, JS unit specs (1290 passed), the two-component and shell e2e specs, and the build pipeline `downport` → `auto_transpile` → `unit` (exit 0). `npm run gates` reports the failures that are pre-existing on `main` (its `check_gates` run is red for the same reasons) and read files this diff does not touch: `check:shared` (sibling repos' `sync-shared.yaml`), `check:mirrors` (the linter's mirrors lag #2777/#2778), `check:changelog` (release 1.144.1 missing in the docs repo), `check:version` (`package.json` 1.144.0 vs `z2ui5_if_app` 1.144.1).
- [x] No manual edits under `src/00/01/`, `src/00/02/` or `src/01/03/` (see AGENTS.md)
- [x] Public API in `src/02/` only changed additively (not touched)

https://claude.ai/code/session_013J521hHjPZZe8sBYH2ByTk


oblomov-dev:
Status on `97d3eec`: the `check_gates` job is red on four gates that are red on `main` (`949081a`, the base of this PR) for the same reasons, and none of them reads a file this diff touches:

- `check:shared` - `.github/shared/sync-shared.yaml` differs from the copies in samples, samples-controls and samples-stack (the sibling repositories lag)
- `check:mirrors` (linter-mirror) - `@abap2ui5/linter` still mirrors `z2ui5_cl_pop_js_loader`, `Z2UI5`, `CLIPBOARD_APP_STATE` and the other names #2777/#2778 removed, and lacks `HASH_BACK` and `setCurrentStep`
- `check:changelog` - release 1.144.1 is in `changelog.txt` but not on the docs repository's changelog page
- `check:version` - `package.json` says 1.144.0, `z2ui5_if_app` says 1.144.1

Each fix lives in another repository or is a release decision (the version bump), so there is nothing to port into this PR; no fix PR exists yet. Every other gate in the job passes, including the repository-wide `eslint .` that failed on the previous push (unused spec variables, fixed in `97d3eec`).

---

